PR #18, opened 2019-08-29
[CE-3884] Add signal listener

CraigLn:


Sources/PubNub/Networking/Endpoints/SubscribeEndpoint.swift
```diff
@@ -102,10 +102,18 @@ public struct TimetokenResponse: Codable, Hashable {
   }
 }
 
+public enum MessageType: Int, Codable {
+  case message = 0
+  case signal = 1
+  case object = 2
+  case presence = 3
```
davidnub: Is `presence` here a placeholder?

CraigLn: Yes

PR #24, opened 2019-10-14
[CE-4088] Listener Token Changes

CraigLn:
User no longer has to manage a token instance when added listeners to subscription event streams.  Subscription listener will now call token.cancel() when it's destroyed. 

Old:
```
let listener = SubscriptionListener()
let token = subscription.add(listener)
```

New:
```
let listener = SubscriptionListener()
subscription.add(listener)
```


.pubnub.yml
```diff
@@ -30,7 +37,7 @@ changelog:
       -
         text: "Fixed HereNow single channel response decoding"
         type: bug
-    date: 2019-XX-XX
+    date: 2019-10-08
```
CraigLn: Is the suggestion to always update the release with the latest commit date or do we want to update this date before/after release with the actual date of release?

davidnub: Good question, not sure we have an exact process. I was just attempting to do it while I was getting ready to release.

PR #28, opened 2019-10-30
Added APIs to add/remove/fetch Message Actions

CraigLn:
Resolves the following tickets: CE-4107, CE-4108, CE-4109, CE-4110, and CE-4111


Sources/PubNub/Networking/Endpoints/Endpoint.swift
```diff
@@ -145,6 +154,7 @@ public enum Endpoint {
 
   // History
   case fetchMessageHistory(channels: [String], max: Int?, start: Timetoken?, end: Timetoken?, includeMeta: Bool)
+  case fetchMessageHistoryWithActions(channel: String, max: Int?, start: Timetoken?, end: Timetoken?, includeMeta: Bool)
```
davidnub: While this is a new API Endpoint in the backend, we had decided to just add a flag in `fetchMessageHistory` titled `includeMessageActions` that would then route to the different endpoint.

Sources/PubNub/PubNub.swift
```diff
@@ -220,6 +220,59 @@ extension PubNub {
...
+  ///   - message: The message to publish
+  ///   - messageAction: The action added to the published message
+  ///   - shouldStore: If true the published message is stored in history.
+  ///   - storeTTL: Set a per message time to live in storage.
+  ///   - meta: Publish extra metadata with the request.
+  ///   - shouldCompress: Whether the message needs to be compressed before transmission
+  ///   - with: Additional network configuration to use on the request
+  ///   - respondOn: The queue the completion handler should be returned on
+  ///   - completion: The async result of the method call
+  public func publishWithMessageAction(
```
davidnub: This seems handy, why would it not just use the usual `publish` method with a parameter for the message action though?

Sources/PubNub/PubNub.swift
```diff
@@ -629,6 +682,47 @@ extension PubNub {
...
+  /// - Parameters:
+  ///   - for: List of channels to fetch history messages from.
+  ///   - max: The max number of messages to retrieve.
+  ///   - start: Time token delimiting the start of time slice (exclusive) to pull messages from.
+  ///   - end: Time token delimiting the end of time slice (inclusive) to pull messages from.
+  ///   - metaInResponse: If `true` the meta properties of messages will be returned as well (if existing).
+  ///   - with: Additional network configuration to use on the request
+  ///   - respondOn: The queue the completion handler should be returned on
+  ///   - completion: The async result of the method call
+  public func fetchMessageHistoryWithMessageActions(
```
davidnub: See comment about `includeMessageActions` flag on the `fetchMessageHistory` function

Sources/PubNub/PubNub.swift
CraigLn: I wasn't sure if there was a practical use-case for this or not, but I needed something like this for integration tests 🤷‍♂

Sources/PubNub/Networking/Endpoints/Endpoint.swift
CraigLn: I can collapse this (on the publish) into the existing method.

Going forward is the preference to add additional method parameters to existing methods when performing similar functions regardless of the difference in server endpoint (history vs history-with-actions) or amount of calls required (publish and add action)?

Sources/PubNub/PubNub.swift
CraigLn: The error being returned would specify the endpoint that was impacted, but if there is not a use-case for it then it can be moved to a test only extension.

Sources/PubNub/PubNub.swift
CraigLn: I can collapse this (on the publish) into the existing method.

Going forward is the preference to add additional method parameters to existing methods when performing similar functions regardless of the difference in server endpoint (history vs history-with-actions) or amount of calls required (publish and add action)?

Edit: Also wanted to add that we shouldn't try and collapse APIs that don't share a common response. History and history-with-actions isn't an example of this, but just more of a food-for-thought.

Sources/PubNub/Networking/Endpoints/Endpoint.swift
davidnub: This is somewhat of a special case because the main usage of the endpoint is almost exactly the same as the one before but includes message actions. The goal here was to make the usage as simple as possible for the developer since that is the main difference.

Sources/PubNub/PubNub.swift
davidnub: Yea that's a good point. I'm not sure we actually need this function.

Sources/PubNub/Networking/Endpoints/Endpoint.swift
CraigLn: Yeah that makes complete sense since everything is the same except that single path component.  I think for more complex cases we want to keep the complexity out of the endpoint, and shift it out elsewhere.

PR #33, opened 2019-11-25
APNS2 Endpoints and Containers

CraigLn:
Includes list/add/remove/remove-all APNS2 device/channel registration
Includes objects to streamline publishing push payloads


Sources/PubNub/Helpers/PubNubAPNSPayload.swift
```diff
@@ -0,0 +1,383 @@
...
+
+  enum CodingKeys: String, CodingKey {
+    case environment
+    case excludedDevices = "excluded_devices"
+    case topic
+  }
+
+  public init(
+    topic: String,
+    environment: PushRouter.Environment = .production,
```
parfeon: At least in Objective-C ticket default environment suggested to be set to “develop”

.pubnub.yml
```diff
@@ -1,9 +1,19 @@
 ---
 name: swift
 scm: github.com/pubnub/swift
-version: "2.2.0"
+version: "2.3.0"
 schema: 1
 changelog:
+  -
+    changes:
+      - test: "Added convience objects for creating push publish messages"
```
polarweasel:
```suggestion
      - test: "Added convenience objects for creating push publish messages"
```

Sources/PubNub/Helpers/PubNubAPNSPayload.swift
CraigLn: From the customer perspective the deployed use-case would be setting it to production, and then develop would be used primarily only during development.

.pubnub.yml
CraigLn: Nice catch will fix

Sources/PubNub/Helpers/PubNubAPNSPayload.swift
davidnub: I think this was a product requirement to set it to `development` by default.

CraigLn: The default can be removed.  My point is that setting the default to development is equivalent to not having a default, because they'll always have to set it to production

PR #41, opened 2020-01-24
[CE-4335] Emit reconnecting event while retrying connection

CraigLn:


Tests/PubNubTests/Integration/SubscriptionIntegrationTests.swift
```diff
@@ -46,7 +46,7 @@ class SubscriptionIntegrationTests: XCTestCase {
     listener.didReceiveSubscription = { event in
       switch event {
       case let .connectionStatusChanged(status):
-        print("Status: \(status)")
+//        print("Status: \(status)")
```
davidnub: Is this left in intentionally?

CraigLn: Yeah it's just for easy debugging of integration tests; not real code

PR #47, opened 2020-05-06
MessageActionEvent improvement to expose the full message response

Ewarrender:
Message actions loaded from history contain a uuid to identify the user. This is an important piece of information useful for either displaying who reacted to a message or showing a selected state if the current user reacted.

This user id is missing from the message action event when received from a subscription. This means it's impossible to identify who the action belongs to. There is also no way to know which channel the message action belongs to.

This pull request exposes the full message response instead of only the 'payload'. MessageActionEvent is now a typealias for MessageResponse<MessageActionSubscribePayload>. To get the MessageActionEventPayload simply use event.payload.data.

@CraigLn

### DIFF
--- a/Sources/PubNub/Networking/Routers/SubscribeRouter.swift
+++ b/Sources/PubNub/Networking/Routers/SubscribeRouter.swift
@@ -583,15 +583,17 @@ public struct MessageActionSubscribePayload: Codable, Hashable {
   let source: String
   let version: String
   let event: MessageActionEventType
-  public let data: MessageActionEvent
+  public let data: MessageActionEventPayload
 }
+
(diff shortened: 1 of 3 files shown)


CraigLn:
Hello, 

Thank you for bringing this to my attention, and I appreciate the PR.  

It looks like this solution changes the structure of the event, and could possibly break functionality for other users.  I've put out a PR as part of our upcoming release that will passively include in the uuid and channel to the existing event structure.  This release should take place sometime today (5/6) or early tomorrow.

Thanks again!

Ewarrender:
Hi @CraigLn,

That's excellent, many thanks for your quick response and implementation!

PR #49, opened 2020-05-18
Add PAM authKey to listChannels(for: channelGroup)

spnkr:
Fixed: If PAM is turned on, `listChannels(for: channelGroup)` returns a 403 because it doesn't include the authToken in the underlying HTTP get.


CraigLn:
Hello, sorry for the delay in response.  

This issue was fixed as part of the 2.5.2 release that is currently available across all supported package managers. 

Thank you!

PR #67, opened 2021-05-29
Use randomized IV by default

parfeon:
## _Crypto_: use randomized IV by default

Crypto by default initialized with `randomizeIV` set to `true` which will encrypt / decrypt data with publish / subscribe / history API calls using randomized initialization vector.

**Warning:** This PR contains **BREAKING CHANGES** which affects how messages are sent with enabled encryption. Randomized IV used by default will make old data inaccessible w/o explicit switch to static IV usage.


CraigLn:
👍  LGTM

CraigLn (approved):

PR #68, opened 2021-06-01
Custom error messages (from functions) will be included inside PubNubError

CraigLn:
This is a minor patch fix to include a custom error message (from Functions) inside the error payload.  

Example usage would be a function aborts a request with a custom message:
```
export default (request) => {
    request.status = 400
    return request.abort('Message dropped')
};
```

The resulting error can be filtered using:
```  
pubnub.publish(channel: "shouldFail", message: payload) { result in
  switch result {
  case let .success(response):
    print("Successful Publish Response: \(response)")
  case let .failure(error as PubNubError):
    if error.reason == .badRequest && error.details.contains("Message dropped") {
      print("Publish Rejected from Function: \(error)")
    } else {
      print("Failed Publish Response: \(error)")
    }
  case let .failure(error):
    print("Failed Publish Response: \(error)")
  }
}
```

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -5,6 +5,8 @@ version: "3.1.3"
 schema: 1
 changelog:
   - changes:
+    - text: "Custom error messages (from functions) will be included inside PubNub Error Details"
+      type: enchancement
     - text: "Fix Coding issue when setting UUIDMetadata objects"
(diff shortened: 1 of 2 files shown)


parfeon (approved):
LGTM

PR #70, opened 2021-06-03
docs: fix change type in .pubnub.yml

parfeon:
## 📕: fix change type in .pubnub.yml

Rename `enchancement` change type to `improvement` which which match requirement of validation schema.


CraigLn (approved):
👍  LGTM

PR #73, opened 2021-07-02
Fixing a couple minor issues  

CraigLn:
* Adding APNS2 Devices will correctly return the devices that were added.
* Per request configuration changes will correct be applied to the outgoing request.

### DIFF
--- a/Sources/PubNub/PubNub.swift
+++ b/Sources/PubNub/PubNub.swift
@@ -438,7 +438,7 @@ extension PubNub {
                            shouldStore: shouldStore,
                            ttl: storeTTL,
                            meta: meta?.codableValue),
-        configuration: configuration
+        configuration: requestConfig.customConfiguration ?? configuration
       )
     } else {
       router = PublishRouter(
@@ -447,7 +447,7 @@ extension PubNub {
                  shouldStore: shouldStore,
                  ttl: storeTTL,
                  meta: meta?.codableValue),
-        configuration: configuration
+        configuration: requestConfig.customConfiguration ?? configuration
       )
     }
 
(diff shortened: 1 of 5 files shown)


Sources/PubNub/PubNub.swift
```diff
@@ -974,11 +1002,16 @@ extension PubNub {
...
     completion: ((Result<[String], Error>) -> Void)?
   ) {
-    route(PushRouter(.manageAPNS(pushToken: deviceToken, environment: environment,
-                                 topic: topic, adding: [], removing: []),
-                     configuration: configuration),
-          responseDecoder: RegisteredPushChannelsResponseDecoder(),
-          custom: requestConfig) { result in
+    route(
+      PushRouter(
+        .manageAPNS(
```
parfeon, line 1007: @CraigLn we are inside of `listAPNSPushChannelRegistrations` and it looks like not really correct to use here `manageAPNS`.
Is it `listPushChannels` which should be used here instead?

parfeon (changes requested):
@CraigLn changes mostly are fine, but one change shown potentially an issue in APNS endpoint usage.

Sources/PubNub/PubNub.swift
CraigLn: This is listing APNS2 registrations exclusively.  The server [API REST docs](https://www.pubnub.com/docs/rest-api#push-notifications-apns-http2) seem to indicate it's the same API as Add/Remove:

parfeon: @CraigLn go it. Then it just confusing enum field name for this type of API.

parfeon (approved):
LGTM :)

Sources/PubNub/PubNub.swift
CraigLn: The router is an internal concept that is meant to map directly to the PubNub server APIs, so they are not meant to be 100% aligned with the public PubNub instance method names.   A new router enum value would typically only be added when the server adds a new API.

PR #77, opened 2021-08-02
Memory leak in subscription loop

parfeon:
## 🐛: memory leak in subscription loop
Fix memory leak caused by circular reference on request in subscription loop.


budgetpreneur (approved):
looks good to me.

PR #78, opened 2021-08-10
PAMv3 implementation

parfeon:
## 🌟 _PAM_: add token parse
Add ability to parse permissions from token provided by servers.

## 🔧 _PAM_: token has higher priority for 'auth'

Use `token` instead of `authKey` (if specified) for query `auth`.

---
**NOTE:** This PR should be merged after #79 (code in this PR should be merged with changes in master after #79 merge to enable release flow).


Sources/PubNub/Networking/HTTPRouter.swift
```diff
@@ -210,7 +212,7 @@ extension HTTPRouter {
       URLQueryItem(name: "uuid", value: configuration.uuid)
     ]
     // Add PAM key if needed
-    if pamVersion != .none, let authKey = configuration.authKey {
+    if pamVersion != .none, let authKey = configuration.token ?? configuration.authKey {
```
CraigLn, line 215: Is there ever a reason why the user would want to be able to use the `authKey` over the `token`, or is this part of the PAM requirements?

Sources/PubNub/PubNub.swift
```diff
@@ -32,13 +32,11 @@ public struct PubNub {
   /// Instance identifier
   public let instanceID: UUID
   /// A copy of the configuration object used for this session
-  public let configuration: PubNubConfiguration
+  public var configuration: PubNubConfiguration
```
CraigLn, line 35: What is the thought of making this a public mutable vs internal/private mutable?  I'm assuming it's solely for the `set(token:)` method below?  

Tokens could be either changed by passing in a custom configuration per-request and/or this could be a private set to allow for the `set(token:)` below if the user doesn't want to manager their own configuration.

Sources/PubNub/PubNubConfiguration.swift
```diff
@@ -116,6 +118,8 @@ public struct PubNubConfiguration: Hashable {
   public var cipherKey: Crypto?
   /// If Access Manager (PAM) is enabled, client will use `authKey` on all requests
   public var authKey: String?
+  /// If Access Manager (PAM) is enabled, client will use  `token` instead of `authKey` on all requests
+  public var token: String?
```
CraigLn, line 122: Is it possible to name this `authToken` to avoid confusion with any other type of `token`? `pamToken` would also work if `authKey` could be changed to `pamKey`

CraigLn (changes requested):

Sources/PubNub/Networking/HTTPRouter.swift
parfeon: Yep:

> Everywhere the SDK supports passing an auth key we should also allow passing a token and give tokens priority over auth keys when both are present

Sources/PubNub/PubNubConfiguration.swift
parfeon: For sure!

CraigLn (approved):
LGTM 👍

Sources/PubNub/Helpers/PAMToken.swift
```diff
@@ -186,12 +102,12 @@ public struct PAMPermission: OptionSet, Codable, Equatable, Hashable {
...
   public static let manage = PAMPermission(rawValue: 1 << 2) // 4
   public static let delete = PAMPermission(rawValue: 1 << 3) // 8
-  public static let create = PAMPermission(rawValue: 1 << 4) // 16
-
-  public static let update: PAMPermission = [PAMPermission.read, PAMPermission.write]
-  public static let crud: PAMPermission = [PAMPermission.create, PAMPermission.update, PAMPermission.delete]
-
-  public static let all: PAMPermission = [PAMPermission.crud, PAMPermission.manage]
+  public static let get = PAMPermission(rawValue: 1 << 5) // 32
+  public static let update = PAMPermission(rawValue: 1 << 6) // 64
```
kleewho, line 106: Why shift? Why not?

```
public static let update = PAMPermission(rawValue: 64)
```

Sources/PubNub/PubNub.swift
```diff
@@ -2034,41 +2030,17 @@ extension PubNub {
...
   public mutating func set(token: String) {
-    tokenStore.set(token: token)
-  }
-
-  /// Stores multiple tokens in the Token Management System for use in API calls.
-  /// - Parameters:
-  ///   - tokens: The list of tokens to add to the Token Management System.
-  public mutating func set(tokens: [String]) {
-    tokenStore.set(tokens: tokens)
+    configuration.authToken = token
```
kleewho, line 2043: What about thread safety? Excuse my lack of any knowledge if the answer is somehow  obvious :)

Sources/PubNub/PubNub.swift
```diff
@@ -32,13 +32,11 @@ public struct PubNub {
   /// Instance identifier
   public let instanceID: UUID
   /// A copy of the configuration object used for this session
-  public let configuration: PubNubConfiguration
+  public private(set) var configuration: PubNubConfiguration
```
kleewho, line 35: I kind of liked the idea of immutable (that's my assumption) PubNubConfiguration and I'd love to keep it that way. I'm actually jealous because it's not immutable in java :(

kleewho (commented):
For what's it worth. I would actually just gut out `TokenStore` and keep there only a way to  `set` and `get` token and keep configuration immutable. But if you're thinking it's stupid idea then go for it

Sources/PubNub/PubNub.swift
parfeon: @kleewho users can’t change it. It can be changed only within this PubNub struct instance.

Sources/PubNub/Helpers/PAMToken.swift
parfeon: @kleewho just sticking with style of existing code.

Sources/PubNub/PubNub.swift
parfeon: This is property, which user won’t change often.
Also this configuration is value type - means that if it is passed somewhere, it will be copy of that structure.

PR #80, opened 2021-08-25
Exclude scripts temporary directory

parfeon:
## Exclude scripts temporary directory

Exclude temporary directory from git.


kleewho (approved):
LGTM

PR #81, opened 2021-09-07
Update .pubnub.yml

crimsonred:


parfeon:
@crimsonred is this change still valid? For some reason it looks like validator doesn't like exactly this new change.

crimsonred:
> @crimsonred is this change still valid? For some reason it looks like validator doesn't like exactly this new change.

Yes it will be when the changes are complete for the validator, converting this to draft for the time being.

CraigLn (approved):

parfeon:
Looks like this change not expected by docs team anymore.

PR #82, opened 2021-09-09
Fix files encryption / decryption

parfeon:
## 🐛 _Files_: fix files encryption / decryption
Fix `bufferTooSmall` and make temporary buffers for files encryption / decryption adaptive in size to provided stream buffer size.


CraigLn (approved):
👍

PR #83, opened 2021-09-10
Don't convert FCM token to HEX

parfeon:
## 🐛 _push_: don't convert FCM token to HEX
Don't perform HEX-encoding of `Data` with FCM token which should be used as-is.


CraigLn (approved):
👍

PR #85, opened 2021-09-16
PubNub SDK 4.1.0

parfeon:
feat(pam): add token parse

Add ability to parse permissions from token provided by servers.

refactor(pam): token has higher priority for 'auth'

Use `token` instead of `authKey` (if specified) for query `auth`.

fix(files): fix files encryption / decryption

Fix `bufferTooSmall` and make temporary buffers for files encryption / decryption adaptive in size to provided stream buffer size.

fix(push): don't convert FCM token to HEX

Don't perform HEX-encoding of `Data` with FCM token which should be used as-is.

feat: add support for components

Make it possible to add component identifiers for requests.

### DIFF
--- a/.pubnub.yml
+++ b/.pubnub.yml
@@ -1,9 +1,22 @@
 ---
 name: swift
 scm: github.com/pubnub/swift
-version: "4.0.2"
+version: "4.1.0"
 schema: 1
 changelog:
+  - date: 2021-09-16
+    version: 4.1.0
+    changes:
+      - type: feature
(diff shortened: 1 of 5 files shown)


parfeon:
@client-engineering-bot release as 4.1.0

CraigLn (approved):

parfeon:
Failure is acceptable, because GitHub Release scripts didn't accepted fact, that artifacts folder can be empty.

PR #97, opened 2022-02-08
Subscribe / presence stateless methods

parfeon:
feat: subscribe / presence stateless methods

Add stateless methods for subscribe / presence API.

### DIFF
--- a/Sources/PubNub/Subscription/SubscriptionSession.swift
+++ b/Sources/PubNub/Subscription/SubscriptionSession.swift
@@ -33,7 +33,7 @@ public class SubscriptionSession {
 
   public let uuid = UUID()
   let longPollingSession: SessionReplaceable
-  internal(set) var configuration: SubscriptionConfiguration
+  var configuration: SubscriptionConfiguration
   let sessionStream: SessionListener
 
   /// PSV2 feature to subscribe with a custom filter expression.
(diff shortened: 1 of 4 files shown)


Sources/PubNub/Subscription/SubscriptionSession.swift
```diff
@@ -33,7 +33,7 @@ public class SubscriptionSession {
 
   public let uuid = UUID()
   let longPollingSession: SessionReplaceable
-  internal(set) var configuration: SubscriptionConfiguration
+  var configuration: SubscriptionConfiguration
```
kleewho, line 36: Does it have to change?

kleewho (commented):
What about `setPresenceState` utility?

Sources/PubNub/Subscription/SubscriptionSession.swift
parfeon: @kleewho this one changed in master already.

parfeon:
@kleewho 

> What about `setPresenceState` utility?

Swift SDK already have `setPresence`. So I'm kind of confusing in why we need to create dupe.

CraigLn:
@parfeon I'm just trying to close out PRs that I'm attached to.  Is this PR still valid?

parfeon:
Yep. This one still valid. Waiting when there will be decision to make release.
Will send it to draft so far.

parfeon:
This has been done in a different way.

PR #105, opened 2022-05-31
Added VSP Modules User, Spaces, and Memberships

CraigLn:
feat: VSP methods and models have been exposed via modules on the PubNub instance

feat: `FlatJSONCodable` replaces `[String: JSONCodableScalar]` inside the VSP models

refactor: PubNub instances has been changed from a `struct` to a `class`

refactor: `HTTPSession` has been made public for easier per-request overrides of the Network Session


PubNubMembership/Sources/Membership+PubNub.swift
```diff
@@ -0,0 +1,518 @@
...
+    completion: @escaping ((Result<(memberships: [PubNubMembership], next: PubNubHashedPage?), Error>) -> Void)
+  )
+
+  /// Fetch all `PubNubMembership` linked to a specific `PubNubSpace.id`
+  ///
+  /// - Parameters:
+  ///   - spaceId: Unique identifier for the `PubNubSpace`.
+  ///   - includeCustom: Should the `PubNubMembership.custom` properties be included in the response
+  ///   - includeSpaceFields: Should the `PubNubUser` properties be included in the response
+  ///   - includeSpaceCustomFields: Should the `PubNubUser.custom` properties be included in the response
```
techwritermat, line 72:
```suggestion
  ///   - includeUserFields: Should the `PubNubUser` properties be included in the response
  ///   - includeUserCustomFields: Should the `PubNubUser.custom` properties be included in the response
```

PubNubMembership/Sources/Membership+PubNub.swift
```diff
@@ -0,0 +1,518 @@
...
+    spaceId: String,
+    includeCustom: Bool,
+    includeUserFields: Bool,
+    includeUserCustomFields: Bool,
+    includeTotalCount: Bool,
+    filter: String?,
+    sort: [PubNub.UserMembershipSort],
+    limit: Int?,
+    page: PubNubHashedPage?,
+    requestConfig: PubNub.RequestConfiguration,
```
techwritermat, line 92: @CraigLn is this parameter optional? If so, shouldn't there be a question mark?

CraigLn: It is not optional, but the implementation uses a default value of `.init()` (which is a shortened way of saying `PubNub.RequestConfiguration()`

jguz-pubnub (approved):

CraigLn:
@parfeon Do you have an estimate on when you'll be able to review these changes?

parfeon:
@CraigLn I should be able to complete it over weekends (need to understand VSP and what it will require from SDK). Don't think, that will find anything critical.

CraigLn:
@parfeon Can you confirm with @KaizenAPI that timeline works with the VSP release deadline, and have him also reach out to the Chat PM to verify everyone agrees on that priority?

parfeon:
@LukGaw215 already reached me out as well for info on this one. Looks like schedule is fine for him.

.pubnub.yml
```diff
@@ -1,9 +1,20 @@
 ---
 name: swift
 scm: github.com/pubnub/swift
-version: "5.1.0"
+version: "6.0.0"
 schema: 1
 changelog:
+  - date: 2022-06-01
```
parfeon, line 7: @CraigLn this information will be duplicated during bot release run. It should be part of PR description.

PubNubMembership/Sources/Membership+PubNub.swift
```diff
@@ -0,0 +1,526 @@
...
+    requestConfig: PubNub.RequestConfiguration = .init(),
+    completion: ((Result<Void, Error>) -> Void)?
+  ) {
+    let router = ObjectsMembershipsRouter(
+      .setMembers(
+        channelMetadataId: spaceId,
+        customFields: nil,
+        totalCount: false,
+        changes: .init(
+          set: users.map { .init(metadataId: $0.user.id, status: $0.status, custom: $0.custom?.flatJSON) },
```
parfeon, line 370: @CraigLn VSP endpoint has been extended and now also allow include `status`?

parfeon (commented):
@CraigLn  Mostly looks good, only one change required and some questions.

PubNubMembership/Sources/Membership+PubNub.swift
CraigLn: Yes, `type` was added to User and Spaces, and `status` was added to User, Spaces, and Memberships.  

Since these fields were added on the Objects V2 those fields will now also exist on Object V2 models and APIs.  There is an include parameter flag for each value (`include=status` and `include=type`) to ensure that field is returned from the APIs, and the SDKs are hardcoded to always include those fields.

parfeon (approved):
LGTM!

@CraigLn maybe for future to decide, but should we do something about deprecated objects API (as I understand, it is moved into `Objects+PubNub.swift`)? Maybe it also should be marked as deprecated?

CraigLn:
> @CraigLn maybe for future to decide, but should we do something about deprecated objects API (as I understand, it is moved into `Objects+PubNub.swift`)? Maybe it also should be marked as deprecated?

I agree.  I asked the question in Slack because I haven't heard anything definitive on when that should happen, but we should do that as soon as we find out.

parfeon:
### VSP Changes:
* Added support for VSP [payloads, methods](https://pubnub.atlassian.net/browse/CLEN-351), and [events](https://pubnub.atlassian.net/browse/CLEN-461).

### Modularity
* Created separate targets within the PubNub SPM dependency as an initial step to modularity. See [ADR PR](https://github.com/pubnub/sdk-specifications/pull/84) for future direction.

### Subscribe changes
* Subscription Listener was split into separate subclasses to allow for module subclassing. See [ADR PR](https://github.com/pubnub/sdk-specifications/pull/84) for more clarification. 
* Subscribe now only decodes the top level Event envelope, and Listener is responsible for decoding event substructure 

### Supporting changes:
* Adds VSP payloads to existing Object v2 models.
* Migrated the Objects v2 code to a separate file extension
* Changed the PubNub instance from a struct to a class
* Created FlatJSONCodable protocol to replace [String: JSONCodableScalar] for VSP objects
* Ran SwiftLint and SwiftFormat commands to uplift codebase
* HTTPSession is now public to make it easier to pass in custom SessionReplaceable's into request parameters
* Made various object public to allow module access

CraigLn:
@client-engineering-bot release as v6.0.0

PR #122, opened 2022-12-12
Use yml validation action

parfeon:
fix: Fix memory leak within Subscription Session

test(validate): use yml validation action

Use specific GitHub action for `.pubnub.yml` validation.


parfeon:
@client-engineering-bot release

PR #125, opened 2022-12-27
Add message type support for: `publish`, `signal`, `history` and `subscribe`

parfeon:
feat(publish): add message type support for `publish` API

Add new publish parameters to specify `messageType` and `spaceId`.

feat(signal): add message type support for `signal` API

Add new signal parameters to specify `messageType` and `spaceId`.

feat(history): add message type support for `history` API

Make it possible to opt-out message type receive for stored messages and get `spaceId` for stored message.

refactor(message-type): refactor `PubNubMessageType` enum

Refactor `PubNubMessageType` enum with ability to hold `custom` message type from user.

BREAKING CHANGES: `PubNubMessageType` enum represented as string instead of int.


parfeon:
New tests won't run, because feature files not ready in test environment (waiting in PR).

Sources/PubNub/PubNub.swift
```diff
@@ -985,7 +1001,7 @@ public extension PubNub {
     for channels: [String],
     includeActions: Bool = false, includeMeta: Bool = false,
     includeUUID: Bool = true, includeMessageType: Bool = true,
-    page: PubNubBoundedPage? = PubNubBoundedPageBase(),
+    includeSpaceId: Bool = false, page: PubNubBoundedPage? = PubNubBoundedPageBase(),
```
jguz-pubnub, line 1004: Nitpick: You can move the `page` parameter to the new line

jguz-pubnub (approved):

Sources/PubNub/PubNub.swift
parfeon: @jguz-pubnub I prefer to have one parameter per-line, if they all doesn't fit into single line. But In this file it looks like two parameters at a line, so I tried to keep with file style.

Tests/PubNubContractTest/Steps/Files/PubNubFilesContractTestSteps.swift
```diff
@@ -156,5 +156,75 @@ public class PubNubFilesContractTestSteps: PubNubContractTestCase {
 
       self.wait(for: [sendFileExpect], timeout: 60.0)
     }
+    
+    When("^I send a file with '(.+)' space id and 'this-is-really-long-message-type-to-be-used-with-publish' message type$") { args, _ in
```
parfeon, line 160: @jguz-pubnub you can have multiple matches in step definition. I would suggest this:
```suggestion
    When("^I send a file with '(.+)' space id and '(.+)' message type$") { args, _ in
```

With code above, you get get last element from `args` and use it for `messageType`.

Tests/PubNubContractTest/Steps/Files/PubNubFilesContractTestSteps.swift
```diff
@@ -156,5 +156,75 @@ public class PubNubFilesContractTestSteps: PubNubContractTestCase {
...
+        case let .failure(error):
+          self.handleResult(result: error)
+        }
+        sendFileExpect.fulfill()
+      }
+
+      self.wait(for: [sendFileExpect], timeout: 60.0)
+    }
+    
+    When("^I send a file with '(.+)' space id and 'ts' message type$") { args, _ in
```
parfeon, line 195: @jguz-pubnub because matched list can dynamically provide you with values, these step looks like redundant if suggested changes will be done.

Tests/PubNubContractTest/Steps/Files/PubNubFilesContractTestSteps.swift
```diff
@@ -156,5 +156,75 @@ public class PubNubFilesContractTestSteps: PubNubContractTestCase {
...
+      guard let data = "test file data".data(using: .utf8) else {
+        XCTAssert(false, "Unable prepare file data")
+        return
+      }
+      
+      self.client.fileURLSession = URLSession(
+        configuration: .default,
+        delegate: self.client.fileSessionManager,
+        delegateQueue: .main
+      )
```
parfeon, line 175: @jguz-pubnub do we need this one to speed up actual upload?

Tests/PubNubContractTest/Steps/Files/PubNubFilesContractTestSteps.swift
jguz-pubnub: Indeed, I checked it again and it's possible to have one template for 3 contracts and there's no need to repeat it:

```
@contract=sendFileWithSpaceIdAndMessageType
@contract=sendFileWithTooShortMessageType
@contract=sendFileWithTooLongMessageType
```

Tests/PubNubContractTest/Steps/Files/PubNubFilesContractTestSteps.swift
jguz-pubnub: I've discovered so far that the default `.pubnub` background session doesn't suppose to work with unit tests, but everything works fine if I run the sample iOS application with `PubNub` changes from this PR and upload a file. This is why I've used the default session configuration in unit tests so far.

parfeon: But file uses `.background` configuration.
What do you mean that it doesn't support?

jguz-pubnub: Sorry, I mislead you. You're right, I also meant `.background` but typed the wrong name in a hurry

parfeon: It looks like a background session is a bad idea for Unit Tests. Thank you, @jguz-pubnub for question on developer forum: https://developer.apple.com/forums/thread/725625

Tests/PubNubContractTest/PubNubContractTestCase.swift
```diff
@@ -91,7 +91,9 @@ let defaultPublishKey = "demo-36"
 
   public var client: PubNub {
     if PubNubContractTestCase.currentClient == nil {
-      PubNubContractTestCase.currentClient = PubNub(configuration: configuration)
+      let client = PubNub(configuration: configuration)
+      client.fileURLSession = URLSession(configuration: .default, delegate: client.fileSessionManager, delegateQueue: .main)
+      PubNubContractTestCase.currentClient = client
```
parfeon, line 103: @jguz-pubnub wait, there is literally parameter in PubNub constructor. Can you please use it? :)

CraigLn (approved):

PR #130, opened 2023-03-16
Change subscribe session settings

parfeon:
Change settings for subscribe URLSession and limit number of concurrent connections per host.

refactor(presence): presence leave on subscribe session

Call presence leave using subscribe URLSession to avoid potential presence race of condition.


jguz-pubnub (approved):

parfeon:
@pubnub-release-bot release

PR #131, opened 2023-03-16
feat/event-engine-poc

jguz-pubnub:
Few assumptions regarding POC:

`EventEngine` is a class with generic parameters (`State`, `Action`, `EffectKind`) that reduces effort and performs some operations related to handling `EffectInvocation` instances, setting a new State for you. This allows you to focus only on providing necessary input in place of generic parameters. These generic parameters are:

1. Provide a `Transition` concrete type. Its requirement is a pure function that returns the new `State` and array of `EffectInvocation` objects for the given `Action`
2. Provide a `Dispatcher` concrete type with `EffectHandlerFactory`. A factory produces an Effect according to provided `EffectInvocation`. This factory returns concrete `Effect` implementation that's tied to the domain described by `EffectKind`

Take a look at the example code I've included in the `Subscribe.swift` file. I assume that everything contained in the `Core` directory should be flexible enough and not require modification. You as a caller should focus only on providing necessary input to the `EventEngine` class. It means that you may also use `Event Engine` for other requirements than Subscribe if you create another instance with different arguments in place of Generics.

### DIFF
--- a/Sources/PubNub/EventEngine/Core/EventEngine.swift
+++ b/Sources/PubNub/EventEngine/Core/EventEngine.swift
@@ -2,7 +2,7 @@
 //  EventEngine.swift
 //
 //  PubNub Real-time Cloud-Hosted Push API and Push Notification Client Frameworks
-//  Copyright © 2019 PubNub Inc.
+//  Copyright © 2023 PubNub Inc.
 //  https://www.pubnub.com/
 //  https://www.pubnub.com/terms
 //
@@ -32,24 +32,25 @@ import Foundation
 /// This class provides a template that allows a caller to specify the concrete types for Event Engine like its State, possible Actions to take, or Effects kind that this Engine produces.
 /// See the `Showcase.swift` file in order to see the example usage.
 ///
-class EventEngine<State: AnyState, Action, EffectKind> {
-  var transition: any TransitionProtocol<State, Action, EffectKind>
-  var dispatcher: any Dispatcher<EffectKind, Action>
+class EventEngine<State: AnyState, Event, EffectKind> {
+  var transition: any TransitionProtocol<State, Event, EffectKind>
+  var dispatcher: any Dispatcher<EffectKind, Event>
   var currentState: State
   
   init(
-    transition: some TransitionProtocol<State, Action, EffectKind>,
-    dispatcher: some Dispatcher<EffectKind, Action>
+    state: State,
+    transition: some TransitionProtocol<State, Event, EffectKind>,
+    dispatcher: some Dispatcher<EffectKind, Event>
   ) {
     self.transition = transition
-    self.currentState = transition.defaultState()
+    self.currentState = state
     self.dispatcher = dispatcher
   }
   
-  func send(action: Action) {
+  func send(event: Event) {
     let result = transition.transition(
       from: currentState,
-      action: action
+      event: event
     )
     
     currentState = result.state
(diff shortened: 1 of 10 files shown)


Sources/PubNub/EventEngine/Core/EventEngine.swift
```diff
@@ -0,0 +1,75 @@
+//
+//  EventEngine.swift
+//
+//  PubNub Real-time Cloud-Hosted Push API and Push Notification Client Frameworks
+//  Copyright © 2019 PubNub Inc.
```
kleewho, line 5: wow. A time machine?

jguz-pubnub: Indeed :) I copied the header comment instead of relying on autogenerated version

Sources/PubNub/EventEngine/Core/EventEngine.swift
```diff
@@ -0,0 +1,75 @@
...
+  var transition: any TransitionProtocol<State, Action, EffectKind>
+  var dispatcher: any Dispatcher<EffectKind, Action>
+  var currentState: State
+  
+  init(
+    transition: some TransitionProtocol<State, Action, EffectKind>,
+    dispatcher: some Dispatcher<EffectKind, Action>
+  ) {
+    self.transition = transition
+    self.currentState = transition.defaultState()
```
are, line 45: I think that the initial `currentState` should be a parameter of the `init` - is there any reason why it comes from the transition protocol?

Sources/PubNub/EventEngine/Core/EventEngine.swift
```diff
@@ -0,0 +1,75 @@
...
+      from: currentState,
+      action: action
+    )
+    
+    currentState = result.state
+    
+    for var invocation in result.invocations {
+      invocation.onCompletion = { [weak self] result in
+        switch result {
+        case .success(let action):
```
are, line 60: I am not sure how I feel about invocations and effects returning an `Action`. What if we want to send multiple actions from an effect?

are:
This looks great! I love how elegant Swift is 😭

Sources/PubNub/EventEngine/Core/EventEngine.swift
jguz-pubnub: There's no strong reason behind that, it can easily be a parameter of the `init` you suggest.

Sources/PubNub/EventEngine/Core/EventEngine.swift
jguz-pubnub: It's not a problem to change it to take an array of `Action`. How did you solve the `Effect` return value in your SDKs? Did you pass the `EventEngine` instance into Effect? This is what I wanted to avoid, but maybe I'm wrong.

jguz-pubnub:
> This looks great! I love how elegant Swift is 😭

Would be happy to welcome you on board! 😉

Sources/PubNub/EventEngine/Core/EventEngine.swift
are: In my implementation, effect handlers have access to `Dependencies` interface, and the `Engine` is one of the dependencies.

jguz-pubnub: Did you hide the `EventEngine` you access from `Dependencies` under some interface? I mean that effect handlers have no knowledge of the whole `EventEngine`. What they know is that's the object they can dispatch an `Event`.

PR #137, opened 2023-05-25
EventEngine

jguz-pubnub:
* Added generic class capable of taking different input for State, Transition, and EffectInvocations
* Added implementation for EffectDispatcher
* State, Transition, and Effects for subscribe loop

### DIFF
--- a/Sources/PubNub/EventEngine/Core/Dispatcher.swift
+++ b/Sources/PubNub/EventEngine/Core/Dispatcher.swift
@@ -1,0 +1,139 @@
+//
+//  Dispatcher.swift
+//
+//  PubNub Real-time Cloud-Hosted Push API and Push Notification Client Frameworks
+//  Copyright © 2023 PubNub Inc.
+//  https://www.pubnub.com/
+//  https://www.pubnub.com/terms
+//
+//  Permission is hereby granted, free of charge, to any person obtaining a copy
+//  of this software and associated documentation files (the "Software"), to deal
+//  in the Software without restriction, including without limitation the rights
+//  to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+//  copies of the Software, and to permit persons to whom the Software is
+//  furnished to do so, subject to the following conditions:
+//
+//  The above copyright notice and this permission notice shall be included in
+//  all copies or substantial portions of the Software.
+//
+//  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+// MARK: - DispatcherListener
+
+struct DispatcherListener<Event> {
+  let onAllInvocationsCompleted: (() -> Void)
+  let onAnyInvocationCompleted: ((String, [Event]) -> Void)
+  
+  init(
+    onAllInvocationsCompleted: @escaping (() -> Void) = {},
+    onAnyInvocationCompleted: @escaping ((String, [Event]) -> Void) = { _, _ in }
+  ) {
+    self.onAllInvocationsCompleted = onAllInvocationsCompleted
+    self.onAnyInvocationCompleted = onAnyInvocationCompleted
+  }
+}
+
+// MARK: - Dispatcher
+
+protocol Dispatcher<Invocation, Event> {
+  associatedtype Invocation: AnyEffectInvocation
+  associatedtype Event
+      
+  func dispatch(invocations: [EffectInvocation<Invocation>], notify listener: DispatcherListener<Event>)
+}
+
+// MARK: - EffectDispatcher
+
+class EffectDispatcher<Invocation: AnyEffectInvocation, Event>: Dispatcher {
+  private let factory: any EffectHandlerFactory<Invocation, Event>
+  private let effectsCache = EffectsCache<Event>()
+  private let effectsRunner = EffectsRunner<Event>()
+    
+  init(factory: some EffectHandlerFactory<Invocation, Event>) {
+    self.factory = factory
+  }
+  
+  func hasPendingEffect(with id: String) -> Bool {
+    effectsCache.hasPendingEffect(with: id)
+  }
+    
+  func dispatch(invocations: [EffectInvocation<Invocation>], notify listener: DispatcherListener<Event>) {
+    let effectsToRun = invocations.compactMap {
+      switch $0 {
+      case .managed(let invocation):
+        return EffectWrapper(id: invocation.id, effect: factory.effect(for: invocation))
+      case .cancel(let id):
+        effectsCache.getEffect(with: id)?.cancelTask(); return nil
+      }
+    }
+    
+    effectsToRun.forEach {
+      effectsCache.put(effect: $0.effect, with: $0.id)
+    }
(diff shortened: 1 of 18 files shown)


Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,325 @@
...
+      ),
+      invocations: [
+        .managed(
+          invocation: .handshakeRequest(
+            channels: newChannels,
+            groups: newGroups
+          )
+        )
+      ]
+    )
```
parfeon, line 120: `SUBSCRIPTION_CHANGED` event also possible on `Receiving` and it's corresponding state variants and should be targeted to `Receiving` state.

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,325 @@
...
+      results = setReceiveReconnectingState(from: state, error: error)
+    case .receiveReconnectGiveUp(let error):
+      results = setReceiveFailedState(from: state, error: error)
+    case .subscriptionChanged(let channels, let groups):
+      results = onSubscriptionChanged(from: state, newChannels: channels, newGroups: groups)
+    case .subscriptionRestored(let channels, let groups, let cursor):
+      results = onSubscriptionRestored(from: state, newChannels: channels, newGroups: groups, cursor: cursor)
+    case .disconnect:
+      results = setStoppedState(from: state)
+    }
```
parfeon, line 90: Am I wrong, or with this implementation, there is no check on whether the current state can handle an event or not?
So, for example, what will happen if state machine is in `ReceivingStopped` and will receive `handshakeFailure` event?

Sources/PubNub/EventEngine/Core/Dispatcher.swift
```diff
@@ -0,0 +1,139 @@
...
+        listeners.forEach { $0.onAnyInvocationCompleted(wrapper.id, results) }
+        group.leave()
+      })
+    }
+    group.notify(
+      queue: DispatchQueue.global(qos: .default),
+      execute: { listeners.forEach { $0.onAllInvocationsCompleted() } }
+    )
+  }
+}
```
parfeon, line 132: Do we need this runner which processes invocations in batches? Will `onAllInvocationsCompleted()` used somewhere?

Sources/PubNub/EventEngine/Core/Dispatcher.swift
```diff
@@ -0,0 +1,139 @@
...
+fileprivate class EffectsCache<Event> {
+  private var managedEffects: Atomic<[String: EffectWrapper<Event>]> = Atomic([:])
+
+  func hasPendingEffect(with id: String) -> Bool {
+    managedEffects.lockedRead { $0[id] } != nil
+  }
+  
+  func put(effect: some EffectHandler<Event>, with id: String) {
+    let existingWrapper = managedEffects.lockedRead { $0[id] }
+    existingWrapper?.effect.cancelTask()
```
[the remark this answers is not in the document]
jguz-pubnub: Right, I will provide you the fixed version for that shortly

Sources/PubNub/EventEngine/Core/Dispatcher.swift
jguz-pubnub: I just wanted to have a callback when all invocations did complete.
But it's not been used so far like you pointed out, so I will remove the entire `EffectsRunner`

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
jguz-pubnub: Not all of them, but some helper methods listed here like [setHandshakeReconnectingState](https://github.com/pubnub/swift/blob/feat/event-engine-effects/Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift#L168-L174) contains `guard` statements in order to check whether the transition should be allowed or not. I could use the same approach for all functions as well. Also, your comment proved to me those functions' names might be extremely confusing. I think it might be better to rename them like `trySetHandshakeReconnectState` or whatever else you come up with.  

I was thinking about the use case with `ReceivingStopped` and `handshakeFailure` you mentioned above. We have roughly 15 events, but only a few of them can be passed to EE due to customer's actions like disconnecting, adding/removing new channels, or restoring subscription. Most of the events passed to EE are the results from our Effects, so my idea was to write unit tests to ensure that every Effect dispatches the expected `Event` back to EE when succeeded/failed. I wanted to repeat the same approach for the hypothetical class like `SubscriptionManager` that holds a reference to EE. Such a class should contain a set of unit tests to verify whether expected events are dispatched to EE as a result of customers' actions like subscribing/unsubscribing/restoring subscription, etc. To sum up, I wanted to rely on unit tests as a safety net in order to prevent writing extra assertions in source code. Do you have any other alternatives/ideas?

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,366 @@
...
+    case .receiveReconnectSuccess(_,_):
+      return state is Subscribe.ReceiveReconnectingState
+    case .receiveReconnectFailure(_):
+      return state is Subscribe.ReceivingState || state is Subscribe.ReceiveReconnectingState
+    case .receiveReconnectGiveUp(_):
+      return state is Subscribe.ReceiveReconnectingState
+    case .subscriptionChanged(_,_):
+      return true
+    case .subscriptionRestored(_, _, _):
+      return true
```
parfeon, line 75: These not possible, at least from `HandshakeStopped` and `ReceiveStopped`.

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,366 @@
...
+    case .disconnect:
+      return true
```
parfeon, line 77: These not possible, at least from `HandshakeFailed` and `ReceiveFailed`.

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,366 @@
...
+  private func canTransition(from state: State, dueTo event: Event) -> Bool {
+    switch event {
+    case .handshakeSucceess(_):
+      return state is Subscribe.HandshakingState
+    case .handshakeFailure(_):
+      return state is Subscribe.HandshakingState
+    case .handshakeReconnectSuccess(_):
+      return state is Subscribe.HandshakeReconnectingState
+    case .handshakeReconnectFailure(_):
+      return state is Subscribe.HandshakingState || state is Subscribe.HandshakeReconnectingState
```
parfeon, line 59: Should `HandshakingState` be in this check? `HANDSHAKE_RECONNECT_FAILURE` triggered only when we are in reconnecting state.

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
```diff
@@ -0,0 +1,366 @@
...
+    case .handshakeReconnectGiveUp(_):
+      return state is Subscribe.HandshakeReconnectingState
+    case .receiveSuccess(_,_):
+      return state is Subscribe.ReceivingState
+    case .receiveFailure(_):
+      return state is Subscribe.ReceivingState
+    case .receiveReconnectSuccess(_,_):
+      return state is Subscribe.ReceiveReconnectingState
+    case .receiveReconnectFailure(_):
+      return state is Subscribe.ReceivingState || state is Subscribe.ReceiveReconnectingState
```
parfeon, line 69: Should `ReceivingState` be in this check? `HANDSHAKE_RECONNECT_FAILURE` triggered only when we are in reconnecting state.

parfeon (changes requested):

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
jguz-pubnub: Right, perhaps the same goes for `HandshakeStopped` and `ReceiveStopped` states.

Sources/PubNub/EventEngine/Subscribe/SubscribeTransition.swift
jguz-pubnub: I allowed them, but I changed implementation of the `onSubscriptionChanged` and `onSubscriptionRestored` helper methods that are called when receiving the `.subscriptionChanged` and `.subscriptionRestored` events respectively. I update the State with a list of provided channels/groups if I detect the transition is either from `HandshakeStopped` or `ReceiveStopped`

PR #138, opened 2023-06-27
feat/event-engine-replacing-old-implementation

jguz-pubnub:
1. Added the generic parameter to `EventEngine` to provide custom input
    - This input can be configuration, listeners, etc, and is passed down to `Effects`
2. Added the new class `SubscribeInput` that keeps currently subscribed channels and channel groups.
    - This class also reduces the effort to compute the final channel and group lists needed to subscribe 
3. Ensuring that sending a new event to `EventEngine` is thread-safe
4. Replacing the old implementation with `EventEngine` and removing items that are not going to be used anymore
5. Ensuring that old unit tests for `SubscribeRouter` work with the new implementation
6. Improving readability in other unit tests
7. Added contract tests


Sources/PubNub/EventEngine/Subscribe/Helpers/SubscribeInput.swift
```diff
@@ -0,0 +1,147 @@
...
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+
+struct SubscribeInput: Equatable {
```
jguz-pubnub, line 30: I don't know if `SubscribeInput` is the best name. This struct aims to keep all channels and channel groups a user is currently subscribed to. The variable of this struct is part of [SubscribeState](https://github.com/pubnub/swift/blob/feat/event-engine-replace/Sources/PubNub/EventEngine/Subscribe/Subscribe.swift#L33) that all concrete state implementations inherit from

parfeon (approved):
LGTM!

PR #140, opened 2023-08-04
feat/event-engine-presence

jguz-pubnub:
EventEngine

* Introduced `isEmpty` variable in `PresenceInput` and `SubscribeInput`
* Added `Equatable` conformance for `Presence.Event` and `Subscribe.Event` to simplify unit tests and assertions
* Unique identifier for Invocation relies on the `id` variable instead of `RawRepresentable<String>`
  * This could be replaced with [Identifiable](https://developer.apple.com/documentation/swift/identifiable) if we start to support iOS 13 and above   
* Implemented state transitions for Presence EE with unit tests
* Implemented all Effects for Presence EE with unit tests
* Reorganized the project's structure and added separate directories for Subscribe and Presence in the test target
* Handling a few transitions through Handshaking according to recent changes in documentation
* Changes related to statuses according to recent changes in documentation
* Renaming: replacing `currentAttempt` with `retryAttempt`
* Added missing unit tests for Subscribe & Presence transitions
* Handling cursor variable and adding missing contract tests for Subscribe EE


Sources/PubNub/EventEngine/Presence/Helpers/PresenceInput.swift
```diff
@@ -0,0 +1,79 @@
...
+  
+  static func +(lhs: PresenceInput, rhs: PresenceInput) -> PresenceInput {
+    var uniqueChannels = lhs.channelsSet
+    var uniqueGroups = lhs.groupsSet
+    
+    rhs.channelsSet.forEach { uniqueChannels.insert($0) }
+    rhs.groupsSet.forEach { uniqueGroups.insert($0) }
+    
+    return PresenceInput(channels: uniqueChannels, groups: uniqueGroups)
+  }
```
parfeon, line 64: @jguz-pubnub can we do it this way? 
```suggestion
  static func +(lhs: PresenceInput, rhs: PresenceInput) -> PresenceInput {
    return PresenceInput(
      channels: lhs.channelsSet.union(rhs.channelsSet), 
      groups: lhs.groupsSet.union(rhs.groupsSet))
  }
```

Sources/PubNub/EventEngine/Presence/Helpers/PresenceInput.swift
```diff
@@ -0,0 +1,79 @@
...
+  
+  static func -(lhs: PresenceInput, rhs: PresenceInput) -> PresenceInput {
+    var uniqueChannels = lhs.channelsSet
+    var uniqueGroups = lhs.groupsSet
+    
+    rhs.channelsSet.forEach { uniqueChannels.remove($0) }
+    rhs.groupsSet.forEach { uniqueGroups.remove($0) }
+    
+    return PresenceInput(channels: uniqueChannels, groups: uniqueGroups)
+  }
```
parfeon, line 74: Same with this one:
```suggestion
  static func -(lhs: PresenceInput, rhs: PresenceInput) -> PresenceInput {
    return PresenceInput(
      channels: lhs.channelsSet.subtract(rhs.channelsSet), 
      groups: lhs.groupsSet.subtract(rhs.groupsSet)
    )
  }
```

Sources/PubNub/EventEngine/Presence/Presence.swift
```diff
@@ -0,0 +1,158 @@
...
+    let input: PresenceInput
+    let error: PubNubError
+  }
+  
+  struct HeartbeatStopped: PresenceState {
+    let input: PresenceInput
+  }
+  
+  struct HeartbeatInactive: PresenceState {
+    let input: PresenceInput = PresenceInput()
```
parfeon, line 81: @jguz-pubnub In `inactive` state according to the graph we don't have any `input`.

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -0,0 +1,220 @@
...
+  typealias State = (any PresenceState)
+  typealias Event = Presence.Event
+  typealias Invocation = Presence.Invocation
+  
+  private func canTransition(from state: State, dueTo event: Event) -> Bool {
+    switch event {
+    case .joined(_,_):
+      return true
+    case .left(_,_):
+      return true
```
parfeon, line 40: `HeartbeatInactive` state can accept only `joined` event for example.

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -0,0 +1,220 @@
...
+    case .heartbeatGiveUp(_):
+      return state is Presence.HeartbeatReconnecting
+    case .timesUp:
+      return state is Presence.HeartbeatCooldown
+    case .leftAll:
+      return true
+    case .disconnect:
+      return true
+    case .reconnect:
+      return state is Presence.HeartbeatStopped || state is Presence.HeartbeatFailed
```
parfeon, line 54: `HeartbeatInactive` state can accept only `joined` event for example.

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+  
+  func onEntry(to state: State) -> [EffectInvocation<Invocation>] {
+    switch state {
+    case is Presence.Heartbeating:
+      return [.managed(.heartbeat(channels: state.channels, groups: state.input.groups))]
```
parfeon, line 61: `Heartbeat` effect is not managed as it is one-shot request.

Sources/PubNub/EventEngine/Presence/Presence.swift
```diff
@@ -0,0 +1,158 @@
...
+}
+
+// MARK: - Presence Effect Invocations
+
+extension Presence {
+  enum Invocation: AnyEffectInvocation {
+    case heartbeat(channels: [String], groups: [String])
+    case leave(channels: [String], groups: [String])
+    case delayedHeartbeat(channels: [String], groups: [String], currentAttempt: Int, error: PubNubError)
+    case scheduleNextHeartbeat(channels: [String], groups: [String])
```
parfeon, line 114: With latest changes we have `wait` instead of `scheduleNextHeartbeat` invocation.

parfeon (changes requested):
There is few questions.

Sources/PubNub/EventEngine/Core/Dispatcher.swift
```diff
@@ -69,9 +69,9 @@ class EffectDispatcher<Invocation: AnyEffectInvocation, Event, Input>: Dispatche
       let effectsToRun = invocations.compactMap {
       switch $0 {
       case .managed(let invocation):
-        return EffectWrapper(id: invocation.rawValue, effect: factory.effect(for: invocation, with: customInput))
+        return EffectWrapper(id: invocation.id, effect: factory.effect(for: invocation, with: customInput))
       case .cancel(let cancelInvocation):
-        effectsCache.getEffect(with: cancelInvocation.rawValue)?.cancelTask(); return nil
+        effectsCache.getEffect(with: cancelInvocation.id)?.cancelTask(); return nil
```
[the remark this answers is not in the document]
jguz-pubnub: Right, I'll introduce something like `regular` invocation

Sources/PubNub/EventEngine/Presence/Helpers/PresenceInput.swift
```diff
@@ -0,0 +1,79 @@
...
+  fileprivate let channelsSet: Set<String>
+  fileprivate let groupsSet: Set<String>
+  
+  var channels: [String] {
+    channelsSet.allObjects.sorted(by: <)
+  }
+  
+  var groups: [String] {
+    groupsSet.allObjects.sorted(by: <)
+  }
```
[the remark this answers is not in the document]
jguz-pubnub: That's only needed for unit tests to validate the `Transition` function. I moved this code into the test target and will provide you the updated version shortly

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -0,0 +1,220 @@
...
+      return [.cancel(.delayedHeartbeat)]
+    default:
+      return []
+    }
+  }
+  
+  func transition(from state: State, event: Event) -> TransitionResult<State, Invocation> {
+    guard canTransition(from: state, dueTo: event) else {
+      return TransitionResult(state: state)
+    }
```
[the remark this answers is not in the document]
jguz-pubnub: Indeed, thx for catching this. I changed it and `canTransition` function is now called from `EE`. An event won't be processed if this function returns false.

Sources/PubNub/EventEngine/Presence/Presence.swift
parfeon: I've missed that `input` is part of `PresenceState` protocol conformance. 

In Rust, each state specifies what data it needs and no “general” requirement because some states shouldn't have channels / groups at all and some has additional data.

So it is fine as it now ;)

Sources/PubNub/EventEngine/Presence/Helpers/PresenceInput.swift
jguz-pubnub: Sure, good catch. I already updated it

Sources/PubNub/EventEngine/Presence/Presence.swift
```diff
@@ -0,0 +1,139 @@
...
+    case wait
+        
+    enum Cancellable: AnyCancellableInvocation {
+      case scheduleNextHeartbeat
+      case delayedHeartbeat
+      
+      var id: String {
+        switch self {
+        case .scheduleNextHeartbeat:
+          return "Presence.ScheduleNextHeartbeat"
```
parfeon, line 119:
```suggestion
        case .wait:
          return "Presence.Wait"
```
Should we use `wait` here?

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -0,0 +1,217 @@
...
+      return [.managed(.delayedHeartbeat(channels: state.channels, groups: state.groups, retryAttempt: state.retryAttempt, error: state.error))]
+    default:
+      return []
+    }
+  }
+  
+  private func onExit(from state: State) -> [EffectInvocation<Invocation>] {
+    switch state {
+    case is Presence.HeartbeatCooldown:
+      return [.cancel(.scheduleNextHeartbeat)]
```
parfeon, line 74: And here `.cancel(.wait)`?

parfeon (approved):
LGTM!

A few questions about naming, but the whole EE is internal, so it shouldn't be a problem.

PR #144, opened 2023-09-25
CryptorModule

jguz-pubnub:
fix(crypto): Improve security of crypto implementation

Improved security of crypto implementation by adding AES-CBC cryptor

feat(crypto): Add CryptorModule

Add CryptorModule that allows configuring SDK to encrypt and decrypt messages


Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+      )
+    }
+    return defaultCryptor.encrypt(data: data).map {
+      if defaultCryptor.id == LegacyCryptor.ID {
+        return $0.data
+      } else {
+        return CryptorHeader.v1(
+          cryptorId: defaultCryptor.id,
+          data: $0.metadata
+        ).asData() + $0.data
```
parfeon, line 89: There are some notes further in PR about `CryptorHeader` that it should represent only information about cryptor, but actual data retrieved and composed here:
```suggestion
        return CryptorHeader.v1(
          cryptorId: defaultCryptor.id,
          dataSize: $0.metadata.count
        ).asData() + $0.metadata + $0.data
```

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+            "Could not find matching Cryptor for \(header.cryptorId()) while decrypting Data. " +
+            "Ensure the corresponding instance is registered in \(String(describing: Self.self))"
+          ]
+        ))
+      }
+      return cryptor.decrypt(
+        data: EncryptedData(
+          metadata: header.metadataIfAny(),
+          data: data.subdata(in: header.length()..<data.count)
+        )
```
parfeon, line 124: If changes to `CryptorHeader` will be done (only size of cryptor-defined data stored) then here will need some changes in metadata computation (`data.subdata()`).

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+    }
+  }
+  
+  /// Encrypts the given `InputStream` object
+  ///
+  /// - Parameters:
+  ///   - stream: Stream to encrypt
+  ///   - contentLength: Content length of encoded stream
+  /// - Returns: A success, storing an ``EncryptedStreamResult`` value if operation succeeds. Otherwise, a failure storing `PubNubError` is returned
+  public func encrypt(stream: InputStream, contentLength: Int) -> Result<EncryptedStreamResult, PubNubError> {
```
parfeon, line 155: Do we really need `EncryptedStreamResult` in response? Can we transparently use `InputStream` both for the source and resulting streams?
I think in tests, I've noticed that we would like to ask user to create `EncryptedStreamResult` to make encryption. What do you think if user will just provide regular `InputStream` and the size of data in it and internally in `CryptoModule.encrypt` method we will do the rest and wrap it into `EncryptedStreamResult`?

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+        data: $0.metadata
+      ) : .none
+      
+      let multipartInputStream = MultipartInputStream(
+        inputStreams: [InputStream(data: header.asData()), $0.stream]
+      )
+      return EncryptedStreamResult(
+        stream: multipartInputStream,
+        contentLength: $0.contentLength + header.length()
+      )
```
parfeon, line 177: I would return an instance of our `MultipartInputStream` here, and it will be transparent for use (`InputStream`).
In send file request configuration, we need the length of the stream (user don't need it because the native interface doesn't have it) and we should be able to get it by casting `InputStream` returned by this method to `MultipartInputStream` and then getting length.

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+  
+  /// Decrypts the given `InputStream` object
+  ///
+  /// - Parameters:
+  ///   - data: A value describing encrypted stream
+  ///   - outputPath: URL where the stream should be decrypted to
+  /// - Returns: A success, storing a decrypted ``EncryptedStreamResult`` value if operation succeeds. Otherwise, a failure storing `PubNubError` is returned
+  @discardableResult
+  public func decrypt(
+    stream streamData: EncryptedStreamResult,
```
parfeon, line 191: As with `encrypt` I would expect user to provide only `InputStream` and size of data in it here.

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -0,0 +1,320 @@
...
+    guard let data = string.data(using: defaultStringEncoding) else {
+      return .failure(PubNubError(
+        .encryptionError,
+        additional: ["Cannot create Data from provided String"]
+      ))
+    }
+    return encrypt(data: data).map {
+      $0.base64EncodedString()
+    }
+  }
```
parfeon, line 306: Did we have this in the past (I mean utility method)? 

Docs for Objective-C contains information about what type should be passed to encryption / decryption and how to create Base64 if resulting data should be sent further.

Sources/PubNub/Helpers/Crypto/Cryptors/LegacyCryptor.swift
```diff
@@ -0,0 +1,184 @@
...
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+//
+
+import Foundation
+import CommonCrypto
+
+/// Provides backward-compatible way of encryption/decryption that matches
+/// deprecated [Crypto](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Helpers/Crypto/Crypto.swift#L32)
```
parfeon, line 32: It will be a dead link because the whole file has been removed (I would expect it still exist in tests, to verify according to Gherkin feature file that data created by `LegacyCryptor` can be decrypted by `Crypto` – to ensure consistency at least in encryption :) )

Sources/PubNub/Helpers/Crypto/Cryptors/LegacyCryptor.swift
```diff
@@ -0,0 +1,184 @@
...
+        options: CCOptions(kCCOptionPKCS7Padding),
+        blockSize: kCCBlockSizeAES128,
+        key: key,
+        initializationVector: ivData,
+        messageData: data
+      )
+      
+      // Join IV and encrypted content when using a random IV
+      return .success(EncryptedData(
+        metadata: ivData,
```
parfeon, line 69: Wouldn't be any issues here if we use random vector, then it will be included both in `metadata` and `data`?
Legacy cryptor doesn't have a header, so there is no metadata for it and all information stored right in the `data` field.

Sources/PubNub/Helpers/Crypto/Cryptors/LegacyCryptor.swift
```diff
@@ -0,0 +1,184 @@
...
+    
+    do {
+      if withRandomIV {
+        iv = data.data.prefix(kCCBlockSizeAES128)
+        cipherText = data.data.suffix(from: kCCBlockSizeAES128)
+      } else {
+        iv = try CryptorVector.fixed.data()
+        cipherText = data.data
+      }
+
```
parfeon, line 92: Should we have some more check here for case, when `cipherText` is empty after we extracted IV from it?

Sources/PubNub/Helpers/Crypto/Header/CryptorHeaderWithinStreamFinder.swift
```diff
@@ -0,0 +1,92 @@
...
+      continuationStream = MultipartInputStream(
+        inputStreams: [
+          InputStream(data: possibleHeaderBytes),
+          stream
+        ]
+      )
+    default:
+      continuationStream = MultipartInputStream(
+        inputStreams: [
+          InputStream(data: possibleHeaderBytes.suffix(from: noOfBytesProcessedByParser)),
```
parfeon, line 52: Can we just use `noOfBytesProcessedByParser` here instead of `suffix` offset computation? Shouldn't this variable store data which hasn't been processed as header?

Sources/PubNub/Helpers/Crypto/Miscellaneous/CryptoInputStream.swift
```diff
@@ -26,11 +26,30 @@
 //
 
 import Foundation
+import CommonCrypto
 
 /// A stream that provides read-only stream functionality while performing crypto operations
 public class CryptoInputStream: InputStream {
   // swiftlint:disable:previous type_body_length
-
+  public struct DataSource {
```
parfeon, line 34: So, right now CrytoInputStream will work only for our cryptors? In future, users also would like to have some help from use in handling stream chunks reading.

Sources/PubNub/Helpers/Crypto/Miscellaneous/CryptoInputStream.swift
```diff
@@ -26,11 +26,30 @@
...
+    let iv: Data
+    let options: CCOptions
+    let cipher: Cipher
+  }
+  public struct Cipher {
+    let algorithm: CCAlgorithm
+    let blockSize: Int
+    
+    func outputSize(from inputBytes: Int) -> Int {
+      inputBytes + (blockSize - (inputBytes % blockSize))
```
parfeon, line 45: CCCrypto actually can tell you what resulting size you will get (depending on from operation type and source data size).

Sources/PubNub/Helpers/Crypto/Miscellaneous/CryptoInputStream.swift
```diff
@@ -57,38 +76,63 @@ public class CryptoInputStream: InputStream {
...
-
-    rawDataLength = contentLength - (operation == .decrypt ? crypto.cipher.blockSize : 0)
-    // The estimated content length is the IV length plus the crypted length
-    estimatedCryptoCount = crypto.cipher.blockSize + crypto.cipher.outputSize(from: rawDataLength)
+    self.includeInitializationVectorInContent = includeInitializationVectorInContent
+    
+    if includeInitializationVectorInContent {
+      rawDataLength = contentLength - (operation == .decrypt ? crypto.cipher.blockSize : 0)
+      // The estimated content length is the IV length plus the crypted length
+      estimatedCryptoCount = crypto.cipher.blockSize + crypto.cipher.outputSize(from: rawDataLength)
```
parfeon, line 97: As mentioned above, `CCCrypto` can help with estimated output size computation.

Sources/PubNub/PubNub.swift
```diff
@@ -1244,38 +1244,6 @@ public extension PubNub {
...
-  func decrypt(data: Data) -> Result<Data, Error> {
-    guard let crypto = configuration.cipherKey else {
-      PubNub.log.error(ErrorDescription.missingCryptoKey)
-      return .failure(CryptoError.invalidKey)
-    }
-
-    return crypto.decrypt(encrypted: data)
-  }
-}
-
```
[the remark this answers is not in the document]
jguz-pubnub: I removed this extension, because it's not `public` (so customers have never seen it) and not used by us internally. Isn't this sufficient?

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
jguz-pubnub: Looks possible. We could return a tuple (under some `typealias` or not) in the response that gathers `InputStream` and its content length. The second part (content length) is needed in order to fill `Content-Length` HTTP header field when uploading a File. This is [how](https://github.com/pubnub/swift/blob/feat/cryptors/Sources/PubNub/Extensions/URLRequest%2BPubNub.swift#L80) I made it in my PR.

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
jguz-pubnub: Yes, I moved those methods from the `Crypto`. They are used internally by us in order to publish or retrieve encoded messages from History. [Here's](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Helpers/Crypto/Crypto.swift#L215-L237) their current implementation in master.

Sources/PubNub/Helpers/Crypto/Miscellaneous/CryptoInputStream.swift
jguz-pubnub: I still preserved this class as `public` so customers can use it. What's changed is the input that this class takes. I decoupled "old" `Crypto` from this class so now customers may create its instances without any dependencies to struct/classes we're going to deprecate.

Sources/PubNub/Helpers/Crypto/Miscellaneous/CryptoInputStream.swift
jguz-pubnub: I moved it from the current master implementation that's available [here](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Helpers/Crypto/Crypto.swift#L142-L147). The reason I moved this code was not to alter/modify `CryptoStream` or `CryptoInputStream` already present before the more it's necessary. 

[CryptoStream](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Helpers/Crypto/CryptoStream.swift) looks like a wrapper for Swift's CommonCrypto operations. In general, I agree with you it's fine to use CommonCrypto's operations directly in the `CryptoInputStream` implementation - since `CryptoInputStream` is already tightly coupled with Swift's CommonCrypto, so IMHO there's nothing that [CryptoStream](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Helpers/Crypto/CryptoStream.swift)  abstracts away. 

I can try to use CCCOperations directly in `CryptoInputStream`, but I'm not sure yet what would be the scope of changes after introducing this idea.

Sources/PubNub/PubNubConfiguration.swift
```diff
@@ -189,7 +189,7 @@ public struct PubNubConfiguration: Hashable {
   /// Specifies the PubNub Subscribe Key to be used when subscribing to a channel
   public var subscribeKey: String
   /// If set, all communication will be encrypted with this key
-  public var cipherKey: Crypto?
+  public var cryptorModule: CryptorModule?
```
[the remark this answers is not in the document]
jguz-pubnub: This should be possible I think, will restore `Crypto` and its fields

Sources/PubNub/Helpers/Streams/MultipartInputStream.swift
```diff
@@ -28,16 +28,18 @@
 import Foundation
 
 /// An `InputStream` that can combine multiple streams into a single stream
-class MultipartInputStream: InputStream {
+public class MultipartInputStream: InputStream {
```
parfeon, line 31: Do we need to make it public? Will be there any benefits for users to using it as `MultipartInputStream`? 

Earlier, this class was created to provide support of multipart-form body, which may consist of sections represented by separate streams with data.

Sources/PubNub/Helpers/Crypto/Header/CryptorHeaderWithinStreamFinder.swift
```diff
@@ -30,33 +30,37 @@ import Foundation
...
+      // There is no CryptorHeader, so all supplied buffer bytes belong to the contents of the File
+      cryptorDefinedData = Data()
+      continuationStream = MultipartInputStream(inputStreams: [InputStream(data: buffer), stream])
+    case .v1(_, let dataLength):
+      // Detects whether it's safe to extract metadata
+      guard headerLength + dataLength < buffer.count else {
+        throw PubNubError(.decryptionFailure, additional: ["Cannot extract metadata for CryptorHeader v1"])
+      }
+      // Extracts Cryptor-defined data from the supplied buffer
+      cryptorDefinedData = buffer.subdata(in: headerLength..<headerLength + dataLength)
```
parfeon, line 54: Just a note, that there can be no cryptor-defined data at all or custom implementation will handle it in a different way.

parfeon (changes requested):

Sources/PubNub/Helpers/Crypto/CryptoError.swift
```diff
@@ -28,6 +28,54 @@
...
+      return hexFrom(Data(hash)).lowercased(with: .current).data(using: .utf8)
+    }
+
+    static func hexFrom(_ data: Data) -> String {
+      let midpoint = data.count / 2
+      return data[..<midpoint].map { String(format: "%02lX", UInt($0)) }.joined()
+    }
+  }
+}
+
```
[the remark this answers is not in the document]
jguz-pubnub: This is the part of "old" `Crypto` that I restored from master to stay backward-compatible. What would you like to move to helpers?

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -101,45 +82,58 @@ public struct CryptorModule {
...
         ))
       }
+      
+      let metadata: Data
+      let contentData: Data
+      
+      switch header {
+      case .none:
+        metadata = Data()
+        contentData = data
```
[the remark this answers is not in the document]
jguz-pubnub: The lack of metadata can be also represented by empty `Data` object. Also, this class provides [isEmpty](https://developer.apple.com/documentation/foundation/data/1780284-isempty) method. I wanted to avoid optional variable that always requires you to write more code or perform additional checks.

Sources/PubNub/Helpers/Streams/MultipartInputStream.swift
jguz-pubnub: I thought it was your idea during the last review 🙂 The whole issue is about setting `Content-Length` when uploading a File, so `CryptorModule` method responsible for stream encryption has to return something more than `InputStream` only. Where do you prefer having encrypted stream length stored?

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
```diff
@@ -151,11 +145,11 @@ public struct CryptorModule {
   /// - Parameters:
   ///   - stream: Stream to encrypt
   ///   - contentLength: Content length of encoded stream
-  /// - Returns: A success, storing an ``EncryptedStreamResult`` value if operation succeeds. Otherwise, a failure storing `PubNubError` is returned
-  public func encrypt(stream: InputStream, contentLength: Int) -> Result<EncryptedStreamResult, PubNubError> {
+  /// - Returns: A success, storing a `MultipartInputStream` value if operation succeeds. Otherwise, a failure storing `PubNubError` is returned
+  public func encrypt(stream: InputStream, contentLength: Int) -> Result<MultipartInputStream, PubNubError> {
```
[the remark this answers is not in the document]
jguz-pubnub: This method is also used for File uploading, so we also need to have encrypted stream length to set proper value for HTTP `Content-Header` field. Where do you store such length?

Sources/PubNub/Helpers/Crypto/Header/CryptorHeaderWithinStreamFinder.swift
jguz-pubnub: I think I will rework that when implementing another Cryptor. Fortunately, that's the internal implementation so it won't break customers in terms of any API. The question is where do you extract `metadata` for `V1` header? Is it inside `CryptorModule` or somewhere else?

Sources/PubNub/Helpers/Crypto/CryptorModule.swift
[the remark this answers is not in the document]
parfeon: OK, makes sense ;)

Sources/PubNub/Helpers/Crypto/Header/CryptorHeaderWithinStreamFinder.swift
parfeon: Since `CryptoModule` after data pre-processing (header extraction) calls to actual cryptor implementations which expect specific types (_encrypt_ expects `Data` and _decrypt_ expects `EncrpyptedData` which contains cryptor-defined portion of data and encrypted data). It is module who extracts `metadata` and puts it into `EncrpyptedData`.

parfeon (approved):
LGTM!

jguz-pubnub:
@pubnub-release-bot release as v6.2.0

PR #147, opened 2023-10-30
License update

jguz-pubnub:
refactor(license): update license information

### DIFF
--- a/PubNubSwift.podspec
+++ b/PubNubSwift.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'PubNubSwift'
-  s.version  = '6.2.1'
+  s.version  = '6.2.2'
   s.homepage = 'https://github.com/pubnub/swift'
   s.documentation_url = 'https://www.pubnub.com/docs/swift-native/pubnub-swift-sdk'
   s.authors = { 'PubNub, Inc.' => 'support@pubnub.com' }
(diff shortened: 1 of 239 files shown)


PubNubSwift.podspec
seba-aln, line 1: just to be sure. You don't have to keep this structure? I mean `:file => 'LICENSE'` ??

jguz-pubnub: I removed it according to [this](https://guides.cocoapods.org/syntax/podspec.html#license) guideline that says:

> Unless the source contains a file named LICENSE.* or LICENCE.*, the path of the license file or the integral text of the notice commonly used for the license type must be specified. If a license file is specified, it either must be without a file extensions or be one of txt, md, or markdown.

Our source tree already contains the `LICENSE` file (without extension) in the root directory, so this should be valid.

seba-aln: Thanks for clarification

parfeon: @jguz-pubnub docs doesn't mention that root may contain file without extension to add automatically. But if you would like to specify it, the name shouldn't have an extension or has one of known extensions.
Or I got it in a wrong way?

jguz-pubnub: Hmm... now I started to have some doubts about that. I could try out the following and check whether `pod spec lint PubNubSwift.podspec` passes or not. Will this solve your worries?

``` s.license = { :type => 'PubNub Software Development Kit License', :file => 'LICENSE' }```

parfeon: @jguz-pubnub this is how I understood words from docs

jguz-pubnub: You're right - our source tree doesn't contain `LICENSE.*` or `LICENCE.*`. It has **LICENSE** file with no extension. So in that case, the path must be specified

parfeon (approved):
LGTM!

jguz-pubnub:
@pubnub-release-bot release as 6.2.2

jguz-pubnub:
@pubnub-release-bot release as 6.2.2

PR #153, opened 2024-02-13
Adding a new Listeners API

jguz-pubnub:
feat(listeners): Adding a new Listeners API


Sources/PubNub/Subscription/SubscriptionSession.swift
```diff
@@ -83,12 +132,33 @@ public class SubscriptionSession {
...
+    channelSubscriptions.forEach { subscription in
+      subscription.subscriptionNames.flatMap { $0 }.forEach {
+        globalChannelSubscriptions[$0] = subscription
+      }
+    }
+    channelGroupSubscriptions.forEach { subscription in
+      subscription.subscriptionNames.flatMap { $0 }.forEach {
+        globalGroupSubscriptions[$0] = subscription
+      }
+    }
```
parfeon, line 161: As I understand, this one is for PubNub client instance itself? 
If subscribing to multiple entities, then it will be more reasonable to create `SubscriptionSet` with `Subscriptions`.

parfeon (commented):
Mostly looks good, but there is a question on ability to have multiple listeners on `Subscription`, `SubscriptionSet` and `PubNub` client. For example, one `Subscription` instance passed in one view between components in it — there is a chance that multiple components would like to have their listener on the same event type.

Sources/PubNub/Events/New/EventEmitter.swift
```diff
@@ -0,0 +1,129 @@
...
+import Foundation
+
+// MARK: - StatusEmitter
+
+/// A protocol for types that emit PubNub status events from the Subscribe loop.
+public protocol StatusEmitter: AnyObject {
+  /// A closure to be called when the connection status changes.
+  var onConnectionStateChange: ((ConnectionStatus) -> Void)? { get set }
+  /// A closure to be called when a subscription error occurs.
+  var onSubscribeError: ((PubNubError) -> Void)? { get set }
```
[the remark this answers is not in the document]
jguz-pubnub: Correct, that's because of the old implementation. You're right, it should be enough to read an underlying error from `ConnectionStatus` enum like `connectionError` and `disconnectedUnexpectedly` and remove `onSubscribeError` closure you pointed out. So what I would suggest is improving existing ConnectionStatus like this:

```
enum ConnectionStatus {
  case connectionError(PubNubError)
  case disconnectedUnexpectedly(PubNubError)
}
```

The reason I haven't done it is that there was `disconnectedUnexpectedly` case introduced a long time ago before EE. Now it's time to change it even if it might be a breaking change. I will run it by the team.

Sources/PubNub/EventEngine/Subscribe/Helpers/SubscribeInput.swift
```diff
@@ -89,52 +104,42 @@ struct SubscribeInput: Equatable {
...
-  ) -> (
-    newInput: SubscribeInput,
-    removedChannels: [PubNubChannel],
-    removedGroups: [PubNubChannel]
-  ) {
+    channels: [PubNubChannel],
+    presenceChannelsOnly: [PubNubChannel],
+    groups: [PubNubChannel],
+    presenceGroupsOnly: [PubNubChannel]
+  ) -> SubscribeInput.RemovingResult {
```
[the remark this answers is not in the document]
jguz-pubnub: I preserved existing behaviors in SDK:

1. You **cannot** subscribe to Presence channel without main channel
2. You **can** unsubscribe from Presence channel only and still keep main channel

The helper method you highlighted is responsible for returning the new channels and groups due to unsubscribing. I can change its signature to be more meaningful:

```
func removing(
  mainChannels: [PubNubChannel]
  presenceOnlyChannels: [PubNubChannel]
  mainGroups: [PubNubChannel],
  presenceOnlyGroups: [PubNubChannel]
)
```

Sources/PubNub/Subscription/Strategy/LegacySubscriptionSessionStrategy.swift
```diff
@@ -308,23 +286,27 @@ class LegacySubscriptionSessionStrategy: SubscriptionSessionStrategy {
...
   // MARK: - Unsubscribe
-
-  func unsubscribe(from channels: [String], and groups: [String], presenceOnly: Bool) {
-    // Update Channel List
+  
+  func unsubscribeFrom(
+    channels: [PubNubChannel],
+    presenceChannelsOnly: [PubNubChannel],
+    groups: [PubNubChannel],
+    presenceGroupsOnly: [PubNubChannel]
```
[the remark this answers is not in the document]
jguz-pubnub: The rules are still the same as I described [in this comment](https://github.com/pubnub/swift/pull/153#discussion_r1493845324). It's a helper method only I can always rename. In general, unsubscribing still works like that:

1. If you unsubscribe from main channel, you should also unsubscribe from its Presence channel
2. If you unsubscribe from Presence channel only, you should still keep main channel in the Subscribe loop

I noticed it's easier to keep these lists in separate parameters when performing next computations.

Sources/PubNub/Subscription/SubscriptionSession.swift
jguz-pubnub: I ran it by the team. You're right that this should be kept in `SubscriptionSet` under the hood. Unfortunately, Swift SDK is the only one that provides functionality like unsubscribing from Presence channels only. And this is what prevents us from using `SubscriptionSet` for global subscriptions. We have no API/method for SubscriptionSet to unsubscribe from Presence channel only. I could deprecate [this method](https://github.com/pubnub/swift/blob/master/Sources/PubNub/PubNub.swift#L399) or even remove `withPresence` parameter, but this decision requires PM input. Can we leave it as it is and fix it in case of further decisions?

jguz-pubnub:
Regarding your comment [here](https://github.com/pubnub/swift/pull/153#pullrequestreview-1885538884), I slightly modified `clone()` methods for `Subscription` and `SubscriptionSet`. Let's suppose that the user would like to pass the same `Subscription` object to another screen, but have different handling for things like `onMessage`, `onPresence`, etc, and not to override any settings from the parent listeners it derives from. Here's the example and it prints "Parent" and "Child" since these two Subscriptions below are independent:

```
let sub = pubnub.channel("someChannel").subscription 
sub.onMessage = { _ in
  print("Parent")
}
sub.subscribe()

let screen = SomeScreen(subscription: sub.clone()) // <====

class SomeScreen: UIViewController {
  let subscription: Subscription
  
  func viewDidLoad() {
    super.viewDidLoad()
    subscription.onMessage = { _ in
       print("Child")
    } 
  }
}
```

Sources/PubNub/EventEngine/Subscribe/Helpers/SubscribeInput.swift
```diff
@@ -120,21 +120,21 @@ struct SubscribeInput: Equatable {
   }
   
   func removing(
-    channels: [PubNubChannel],
+    mainChannels: [PubNubChannel],
     presenceChannelsOnly: [PubNubChannel],
-    groups: [PubNubChannel],
+    mainGroups: [PubNubChannel],
     presenceGroupsOnly: [PubNubChannel]
```
parfeon, line 126: This could be as simple as:
```swift
  func removing(
    channels: [PubNubChannel],
    presenceChannels: [PubNubChannel],
    groups: [PubNubChannel],
    presenceGroups: [PubNubChannel]
```
Also, should there be `PubNubChannelGroup` instead of `PubNubChannel`?

Sources/PubNub/Events/New/Subscription.swift
```diff
@@ -165,35 +169,31 @@ extension Subscription: SubscribeMessagesReceiver {
...
+    guard entityName.hasSuffix(".*") else {
+      return entityName == string
     }
-    let pattern = "^" + self + "$"
-    let predicate = NSPredicate(format: "SELF MATCHES %@", pattern)
-
-    return predicate.evaluate(with: string)
+    if let firstIndex = entityName.lastIndex(of: "."), let secondIndex = string.lastIndex(of: ".") {
+      return entityName.prefix(upTo: firstIndex) == string.prefix(upTo: secondIndex)
+    }
```
parfeon, line 196: Isn't the "original" `NSPredicate` basically matched the same what is done on the line `#192`?

Sources/PubNub/Subscription/Strategy/SubscriptionSessionStrategy.swift
```diff
@@ -27,9 +27,9 @@ protocol SubscriptionSessionStrategy: AnyObject {
     at cursor: SubscribeCursor?
   )
   func unsubscribeFrom(
-    channels: [PubNubChannel],
+    mainChannels: [PubNubChannel],
     presenceChannelsOnly: [PubNubChannel],
-    groups: [PubNubChannel],
+    mainGroups: [PubNubChannel],
```
parfeon, line 32: Should we use `PubNubChannelGroup` here?

parfeon (approved):
There is some naming questions and one about whether proper type used in method signature or not.

About multiple listeners of the same type on single `Subscription` / `SubscriptionSet` is possible only through `clone()` utilization and impossible to have multiple `onMessage` listeners on the same instance.

Sources/PubNub/Events/New/Subscription.swift
jguz-pubnub: Yeah, it should. It compares two String up to the last occurrence of `.`, assuming both contain the `.` character. Let's suppose these scenarios:

1. The message came from `channel.item.x` and the underlying entity's subscription is `channel.item.*`:
    -  According to the line 195, I will compare `channel.item` with `channel.item` ✅ 
2. The message came from `a.b.c` and the underlying entity's subscription is `channel.item.*`
    - I will compare `a.b` with `channel.item` 🔴

Sources/PubNub/EventEngine/Subscribe/Helpers/SubscribeInput.swift
jguz-pubnub: Unfortunately, there's one struct representing channel or channel group. This struct was even exposed as public and its instances are parameters for[ local events](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Events/Subscription/SubscriptionStream.swift#L17-L21). The way we may consider reworking it is:

1. Removing `[PubNubChannel]` from these local events and replace it with other type or even raw `[String]`
2. Provide different type representing a channel group. This type should be used internally for our implementation

This might be a breaking change for someone who uses local events. On the other hand, we don't plan to support them anymore. So if I get the approval, I can introduce these changes after prior consultation with you.

Sources/PubNub/Events/New/Subscription.swift
parfeon: Ok, I finally got it. 
I was totally fine with `extension` for `String` - it looked better.

I probably would use `payload.subscription ?? payload.channel` to match both `channel` and `channelGroup` subscription type. If a subscription has been created for wildcard subscription, then the `subscription` field will hold the same name (if an event for one of the channels from it) and will be no need for pattern match (just plain `==`). The only thing which may require some attention is presence channels / channel group names.

techwritermat (commented):
enable event engine please :)

PR #155, opened 2024-02-23
Upgrading deployment targets

jguz-pubnub:
- Upgarding deployment targets to minimum deployment targets suggested by `Xcode 15.3`:
  - `iOS 12.0` and above
  - `tvOS 12.0` and above
  - `watchOS 4.0` and above
  - `macOS 10.13` and above
- Updating `README.md`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -14,10 +14,10 @@ import PackageDescription
 let package = Package(
   name: "PubNub",
   platforms: [
-    .iOS(.v9),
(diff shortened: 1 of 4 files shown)


parfeon (changes requested):
What about `Package.swift`?

parfeon (approved):
LGTM!

kazydek (approved):

PR #157, opened 2024-03-05
Updating instructions for fetching PubNub SDK via SPM

jguz-pubnub:
I missed that the instruction for fetching PubNub SDK via Swift Package Manager is also outdated. This PR fixes it.


parfeon (approved):
LGTM!

PR #160, opened 2024-03-13
Subscribing to the same channel via PubNub class causes disconnecting

jguz-pubnub:
fix(subscribe): prevent from disconnecting while subscribing to the channel that was already subscribed


parfeon (approved):
LGTM!

jguz-pubnub:
@pubnub-release-bot release as 7.1.0

PR #180, opened 2024-06-17
Calling .leave request when unsubscribing

jguz-pubnub:
fix(presence): calling .leave request in missing cases


Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
```diff
@@ -26,15 +26,15 @@ class PresenceTransition: TransitionProtocol {
...
+      return true
     case .heartbeatSuccess:
       return state is Presence.Heartbeating
     case .heartbeatFailed:
       return state is Presence.Heartbeating
     case .timesUp:
       return state is Presence.HeartbeatCooldown
     case .leftAll:
-      return !(state is Presence.HeartbeatInactive)
+      return true
```
parfeon, line 37: So, we want to transit even if heartbeat loop is not active?

parfeon (approved):
LGTM!

Sources/PubNub/EventEngine/Presence/PresenceTransition.swift
jguz-pubnub: Yes. You're right, the heartbeat loop is inactive if the user subscribes to some channel without Presence. However, we still have to trigger `.leave` invocation as a part of the unsubscribe cleanup

jguz-pubnub:
@pubnub-release-bot release as 7.3.1

PR #181, opened 2024-07-15
Refactor PubNubObjC creation and config access

wkal-pubnub:


Sources/PubNub/KMM/PubNubObjC.swift
```diff
@@ -65,3 +76,25 @@ public extension PubNubObjC {
...
+
+@objc
+public class PubNubConfigurationObjC: NSObject {
+    let configuration: PubNubConfiguration
+      
+    public init(configuration: PubNubConfiguration) {
+      self.configuration = configuration
+    }
+    
+    @objc
```
jguz-pubnub, line 90: You can shorten the whole syntax a bit:

```
@objc
public var userId: String {
   configuration.userId
}
```

Sources/PubNub/KMM/PubNubObjC.swift
```diff
@@ -65,3 +76,25 @@ public extension PubNubObjC {
...
+    public var userId: String {
+        @objc get { return configuration.userId }
+    }
+    
+    @objc
+    public var authKey: String? {
```
jguz-pubnub, line 96: The same here:

```
@objc
public var authKey: String? {
    configuration.authKey
}
```

Sources/PubNub/KMM/PubNubObjC.swift
wkal-pubnub: done

Sources/PubNub/KMM/PubNubObjC.swift
wkal-pubnub: done

jguz-pubnub (approved):

PR #185, opened 2024-08-14
Upgrading PubNub Swift SDK

jguz-pubnub:
### Breaking changes:

- Module name has been replaced with `PubNubSDK` due to compiler error when a public type has the same name as a module. This is the known issue reported in https://github.com/swiftlang/swift with `56573`. As a consequence, customers have to replace `import PubNub` with `import PubNubSDK` in their Swift code
- `ConnectionStatus.connecting` and `ConnectionStatus.reconnecting` are no longer supported. This version introduces the new `.subscriptionChanged` connection status indicating the SDK subscribed with new channels/channel groups, fired every time the channel/channel group mix changed, since the initial connection

### Features:

- Adding possibility to attach listeners to `Subscription` and `SubscriptionSet`

### Fixes:

- Ensuring that listeners are internally stored as weak references


Sources/PubNub/EventEngine/Subscribe/Effects/EmitMessagesEffect.swift
```diff
@@ -37,14 +37,14 @@ class MessageCache {
 struct EmitMessagesEffect: EffectHandler {
   let messages: [SubscribeMessagePayload]
   let cursor: SubscribeCursor
-  let listeners: [BaseSubscriptionListener]
+  let listeners: WeakSet<BaseSubscriptionListener>
```
wkal-pubnub, line 40: what happens if someone adds a listener to a subscription, but doesn't keep a reference to it otherwise?

for example, if I did this in Kotlin with weak references:
```
subscription.addListener ( object : EventListener { override fun onMessage() {...} } )
```

then after a while, I would stop receiving events because no one holds a strong reference to the listener, so it gets garbage collected

jguz-pubnub: It looks like those names are extremely confusing. In fact, the `listeners` variable keeps instances to all subscriptions. This could be either `Subscription`, `SubscriptionSet`, or the item like `SubscriptionListener` which comes from the "old" API that we still have to support. So perhaps the best idea would be replacing variable's name from the line 40 with `subscriptions` or something less confusing.

Regarding your code snippet, classes like `Subscription` and `SubscriptionSet` internally store their listeners as **strong references**. Thus, you'll be able to receive events

PR #193, opened 2024-10-23
Update Ruby gems and the default runner group

jguz-pubnub:
chore(project): update Ruby gems and the default runner group


parfeon (approved):
LGTM!

PR #194, opened 2024-11-22
Add support for the `type` field in PubNubMembershipMetadata

jguz-pubnub:
feat(membership): add support for the `type` field in `PubNubMembershipMetadata`

feat(app-context): deprecate methods containing `uuid` parameter,  introduce new versions

feat(app-context): deprecate methods containing `include` parameter of `Bool` type, introduce new versions

feat(app-context): deprecate `remove(channel:custom:completion:)`, introduce new version 

feat(app-context): replace `PubNubUUIDMetadata` with `PubNubUserMetadata`

feat(app-context): replace `PubNubUUIDMetadataBase` with `PubNubUserMetadataBase`

feat(app-context): add an alias for backward compatibility with the customer code using `PubNubUUIDMetadata`

feat(app-context): add an alias for backward compatibility with the customer code using `PubNubUUIDMetadataBase`

feat(listeners): replace `PubNubUUIDMetadataChangeset` with `PubNubUserMetadataChangeset`

feat(listeners): add an alias for backward compatibility with the customer code using `PubNubUUIDMetadataChangeset`


PubNubUser/Sources/PubNubUser.swift
```diff
@@ -139,7 +139,7 @@ extension PubNubUser: Codable {
 
 // MARK: Object v2 Migration
 
-public extension PubNubUUIDMetadata {
+public extension PubNubUserMetadata {
   /// Converts Object V2 UUID Metadata to a Space entity
   ///
   /// - returns: The `PubNubUser` built from the Object V2 data
```
parfeon, line 145: Do we want to rename object in comments as well?

PubNubUser/Sources/User+PubNub.swift
```diff
@@ -205,7 +205,7 @@ public extension PubNubUserInterface {
   ) {
     let router = ObjectsUUIDRouter(
       .all(
-        customFields: includeCustom,
+        include: PubNub.IncludeFields(custom: includeCustom).userIncludeFields,
```
parfeon, line 208: Will it make more sense to have something like this:
```suggestion
        include: PubNub.UserIncludeFields(custom: includeCustom),
```

Sources/PubNub/APIs/Objects+PubNub.swift
```diff
@@ -90,18 +104,20 @@ public extension PubNub {
...
       self.statusField = statusField
+      self.typeField = typeField
       self.channelCustomFields = channelCustomFields
       self.channelTypeField = channelTypeField
       self.channelStatusField = channelStatusField
       self.totalCount = totalCount
     }
 
-    var customIncludes: [ObjectsMembershipsRouter.Include]? {
+    var includeFields: [ObjectsMembershipsRouter.Include]? {
```
parfeon, line 114: So it had a wrong name for all this time?

Sources/PubNub/APIs/Objects+PubNub.swift
```diff
@@ -251,95 +310,226 @@ public extension PubNub {
...
+  ///   - completion: The async `Result` of the method call
+  ///     - **Success**: The `PubNubUserMetadata` object belonging to the identifier
+  ///     - **Failure**: An `Error` describing the failure
+  func fetchUserMetadata(
+    _ metadataId: String?,
+    include: IncludeFields = IncludeFields(),
+    custom requestConfig: RequestConfiguration = RequestConfiguration(),
+    completion: ((Result<PubNubUserMetadata, Error>) -> Void)?
   ) {
     let router = ObjectsUUIDRouter(
```
parfeon, line 425: Do we want to rename router as well?

Sources/PubNub/APIs/Objects+PubNub.swift
```diff
@@ -474,9 +731,11 @@ public extension PubNub {
       configuration: requestConfig.customConfiguration ?? configuration
     )
 
-    route(router,
-          responseDecoder: GenericServiceResponseDecoder(),
-          custom: requestConfig) { result in
+    route(
```
parfeon, line 734: If we have explicit method names to set channel / user metadata, do we want to have the same for `remove`?

Sources/PubNub/Models/PubNubMembershipMetadata.swift
```diff
@@ -61,62 +69,112 @@ public extension PubNubMembershipMetadata {
...
+  public var updated: Date?
+  /// The caching identifier for the object
+  public var eTag: String?
 
-  var concreteUUID: PubNubUUIDMetadataBase?
-  public var uuid: PubNubUUIDMetadata? {
-    get { concreteUUID }
-    set {
-      concreteUUID = try? newValue?.transcode()
+  var concreteUUID: PubNubUserMetadataBase?
```
parfeon, line 86: Do we want to rename this to the `concreteUser`?

Sources/PubNub/Networking/Routers/ObjectsChannelRouter.swift
```diff
@@ -10,13 +10,17 @@
...
+}
+
 public struct ObjectsChannelRouter: HTTPRouter {
   public enum Endpoint: CustomStringConvertible {
-    case all(
-      customFields: Bool, totalCount: Bool, filter: String?, sort: [String], limit: Int?, start: String?, end: String?
-    )
-    case fetch(metadataId: String, customFields: Bool)
-    case set(metadata: PubNubChannelMetadata, customFields: Bool)
+    case all(include: [Include]?, totalCount: Bool, filter: String?, sort: [String], limit: Int?, start: String?, end: String?)
```
parfeon, line 21: Isn't `totalCount` part of inclusion flags? Maybe it will make sense to merge into `Include`?

Sources/PubNub/Networking/Routers/ObjectsChannelRouter.swift
```diff
@@ -100,14 +98,14 @@ public struct ObjectsChannelRouter: HTTPRouter {
       query.appendIfPresent(key: .filter, value: filter)
       query.appendIfNotEmpty(key: .sort, value: sort)
       query.appendIfPresent(key: .limit, value: limit?.description)
-      query.appendIfPresent(key: .include, value: Include.includes(custom: customFields).csvString)
+      query.appendIfPresent(key: .include, value: Include.includes(from: customFields))
```
parfeon, line 101: `includeFields` instead of `customFields` because it is not really custom?

parfeon (commented):
Mostly looks good! :)
Just a few questions.

PubNubUser/Sources/User+PubNub.swift
jguz-pubnub: I agree, it makes sense. I'll introduce it and let you know when this becomes available

Sources/PubNub/APIs/Objects+PubNub.swift
jguz-pubnub: Yeah, I would say the previous name was quite confusing

Sources/PubNub/APIs/Objects+PubNub.swift
jguz-pubnub: 1. Regarding removing User metadata, here's the method we have so far - which I deprecated in this PR:

```
@available(*, deprecated, renamed: "removeUserMetadata(_:custom:completion:)")
func remove(
   uuid metadataId: String?,
   custom requestConfig: RequestConfiguration = RequestConfiguration(),
   completion: ((Result<String, Error>) -> Void)?
)
```
From now, the recommended method for removing User metadata would be:

```
func removeUserMetadata(
   _ metadataId: String?,
   custom requestConfig: RequestConfiguration = RequestConfiguration(),
   completion: ((Result<String, Error>) -> Void)?
)
```

2. Regarding removing Channel metadata, nothing has changed in the scope of this PR so the customers will still use the same method as previously:

```
func remove(
   channel metadataId: String,
   custom requestConfig: RequestConfiguration = RequestConfiguration(),
   completion: ((Result<String, Error>) -> Void)?
)
```

Do you think is it OK? Or would you like to introduce the new method for removing Channel metadata and slightly change its name?

```
func removeChannelMetadata(
   _ metadataId: String,
   custom requestConfig: RequestConfiguration = RequestConfiguration(),
   completion: ((Result<String, Error>) -> Void)?
)
```

parfeon: It is just kind of weird that one metadata type `remove` strongly `emphasized` and for another we would know only if review argument label names.

Sources/PubNub/Networking/Routers/ObjectsChannelRouter.swift
parfeon: Ok, let's keep it as is for now.

Sources/PubNub/APIs/Objects+PubNub.swift
jguz-pubnub: Am I right? Would you prefer this method? If so, I could introduce it:

```
func removeChannelMetadata(
   _ metadataId: String,
   custom requestConfig: RequestConfiguration = RequestConfiguration(),
   completion: ((Result<String, Error>) -> Void)?
)
```

parfeon (approved):
LGTM!

PR #198, opened 2025-01-10
Fix the issue of passing invalid values to the `publish` method in the KMP layer (internal usage)

jguz-pubnub:
fix(kmp): fix the issue of passing an invalid value for the `storeTTL` parameter in the KMP layer (internal use)
fix(kmp): handle `shouldCompress` parameter in the KMP layer  (internal use)


jguz-pubnub:
@pubnub-release-bot release as 8.2.4

PR #199, opened 2025-01-15
Add synchronized access to EventEngine's `state` property

jguz-pubnub:
fix(subscribe): fix the crash issue caused by multiple accesses to the `state` property in multithreaded scenarios


marcin-cebo (approved):

jguz-pubnub:
@pubnub-release-bot release as 8.2.5

PR #204, opened 2025-03-11
Include endpoint error message in PubNubError's details property

jguz-pubnub:
fix(error-handling): include endpoint error message in PubNubError's `details` property

fix(logger): defer computing the textual representation of EventEngine's current and previous states until needed


Sources/PubNub/Errors/ErrorDescription.swift
```diff
@@ -197,7 +197,7 @@ extension PubNubError.Reason: CustomStringConvertible, LocalizedError {
     case .unauthorized:
       return "Access was denied due to insufficient authentication/authorization"
     case .forbidden:
-      return "Authorization key is missing or does not have the permissions required to perform this operation"
+      return "No permissions required to perform this operation"
```
parfeon, line 200: This doesn't sound right. It literally says that there is no special permissions required, when we get at 403.

Sources/PubNub/Networking/Response/ResponseOperator.swift
```diff
@@ -81,12 +81,23 @@ public extension ResponseDecoder {
       GenericServicePayloadResponse.self,
       from: data
     )
+
+    let fullDetails: [ErrorDetail]? = if let payload = generalErrorPayload {
+      [ErrorDetail(
+        message: payload.message.rawValue,
```
parfeon, line 87: Sometimes error message stored inside of `error` key. But our newest versions of endpoints stick to the new format.

parfeon (approved):
There is wording in error which may need to be changed.

Sources/PubNub/Errors/ErrorDescription.swift
jguz-pubnub: I will change it to: `"Insufficient permissions to perform this operation`"

Sources/PubNub/Networking/Response/ResponseOperator.swift
jguz-pubnub: Yes, you're right. [Here's](https://github.com/pubnub/swift/blob/master/Sources/PubNub/Networking/Response/GenericServicePayloadResponse.swift#L250) the full handling for different error message response structures. And all in all, it's parsed into `EndpointResponseMessage`

jguz-pubnub:
@pubnub-release-bot release as 9.0.1

PR #206, opened 2025-05-12
Enhance file encryption with stream metadata

jguz-pubnub:
feat(crypto-module): return stream metadata when encrypting a file

The `encrypt(stream:contentLength:)` method in `CryptoModule` now returns an `EncryptedStreamResult` containing both the encrypted stream and its total content length

```sw
// Before
if case .success(let result) = cryptoModule.encrypt(stream: fileStream, contentLength: fileSize) {
  // The `result` variable is of type `InputStream` representing an encoded stream.
  // There's no information about the total length of the encoded stream
}

// Now
if case .success(let result) = cryptoModule.encrypt(stream: fileStream, contentLength: fileSize) {
  // The result variable is of type `EncryptedStreamResult`, containing both the encrypted stream 
  // and its content length:
  let stream = result.stream 
  let totalLength = result.contentLength
}
```


Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -273,6 +275,48 @@ public struct CryptoModule {
...
+  /// Decrypts the stream from the given URL and writes it to the output path
+  ///
+  /// - Parameters:
+  ///   - url: URL of the encrypted stream
+  ///   - outputPath: Path to write the decrypted stream
+  /// - Returns:
+  ///  - **Success**: A decrypted `InputStream` object
+  ///  - **Failure**: `PubNubError` describing the reason of failure
+  @discardableResult
+  public func decryptStream(from url: URL, to outputPath: URL) -> Result<InputStream, PubNubError> {
```
marcin-cebo, line 287: Is there old decrypt Method that The `result` variable is of type `InputStream`. Should it get deprecation message?

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -273,6 +275,48 @@ public struct CryptoModule {
...
+    )
+
+    guard let inputStream = InputStream(url: url) else {
+      PubNub.log.debug(
+        "Cannot create InputStream from \(url). Ensure that the file exists at the specified path",
+        category: .crypto
+      )
+      return .failure(PubNubError(
+        .decryptionFailure,
+        additional: ["File doesn't exists at \(url) path"]
```
marcin-cebo, line 300: Shouldn't be "exist" ?

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -273,6 +275,48 @@ public struct CryptoModule {
...
+      ))
+    }
+
+    let streamDecryptionResult = performStreamDecryption(
+      stream: inputStream,
+      contentLength: url.sizeOf,
+      to: outputPath
+    )
+
+    switch streamDecryptionResult {
```
marcin-cebo, line 310: Shouldn't be like this:
```
switch streamDecryptionResult {
case let .success(decryptedStream):
  PubNub.log.debug("File decrypted successfully", category: .crypto)
  return .success(decryptedStream)

case let .failure(error):
  PubNub.log.debug("Decryption of file failed due to \(error)", category: .crypto)
  return .failure(error)
}
```

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -273,6 +275,48 @@ public struct CryptoModule {
...
+        additional: ["File doesn't exists at \(url) path"]
+      ))
+    }
+
+    let streamDecryptionResult = performStreamDecryption(
+      stream: inputStream,
+      contentLength: url.sizeOf,
+      to: outputPath
+    )
+
```
marcin-cebo, line 309: Shouldn't we add:
```
 inputStream.open()
  defer { inputStream.close() }
```

The thing is that you’re creating the InputStream from disk, but you never open it (or close it on all paths). A simple inputStream.open()/defer { inputStream.close() } around your decrypt call will ensure the file handle is managed correctly.

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -273,6 +275,48 @@ public struct CryptoModule {
...
+      return .failure(PubNubError(
+        .decryptionFailure,
+        additional: ["File doesn't exists at \(url) path"]
+      ))
+    }
+
+    let streamDecryptionResult = performStreamDecryption(
+      stream: inputStream,
+      contentLength: url.sizeOf,
+      to: outputPath
```
marcin-cebo, line 307: What happens if a file already exists?

marcin-cebo (approved):

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
jguz-pubnub: Sure, this could also be like this. The reason I haven't done it this way was to keep one return statement instead of two. There's no difference in general, so I'll change it to your suggestion

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
jguz-pubnub: This is the URLSession object that will open and close the stream when uploading it to the remote. It's not the CryptoModule that will manage it

Sources/PubNub/Helpers/Crypto/CryptoModule.swift
jguz-pubnub: We would like to treat this method as a convenient method to already existing method responsible for decoding a stream. It's up to you as a customer which one you would like to pick.

It's likely we will deprecate existing `decrypt(stream: contentLength: to:)` method. However, I don't wanna do it in a hurry and consider first all the pros and cons before doing it

jguz-pubnub:
@pubnub-release-bot release as 9.1.0

PR #207, opened 2025-05-15
Add a convenience method for encrypting a stream

jguz-pubnub:
feat(crypto-module): add a convenience `encryptStream(from:)` method in CryptoModule

The `encryptStream(from:)` convenience method is designed to encrypt a local file. It eliminates the need for the caller to provide an `InputStream` and its content length - these are handled internally. You only need to pass a local file URL


Sources/PubNub/Helpers/Crypto/CryptoModule.swift
```diff
@@ -317,8 +357,7 @@ public struct CryptoModule {
     return streamDecryptionResult
   }
 
-  @discardableResult
```
jguz-pubnub, line 320: I accidentally marked this helper method as public. This PR also fixes it

parfeon (approved):
LGTM!

PR #219, opened 2025-07-31
Fix the issue with recursive locking when disposing of a SubscriptionSet

jguz-pubnub:
fix(subscription-set): fix the issue with recursive locking introduced in 9.3.0 when disposing a SubscriptionSet


marcin-cebo (approved):

jguz-pubnub:
@pubnub-release-bot release as 9.3.1

PR #221, opened 2025-08-07
Fix APM library crash (when present) by centralizing HTTPSession cleanup

jguz-pubnub:
fix(presence): fix APM library crash (when present) by centralizing `HTTPSession` cleanup

APM Libraries Crash Fix: Removed explicit deinit methods from `PresenceEffectFactory` and `SubscribeEffectFactory` to prevent race conditions with Application Performance Monitoring libraries. These factories now rely entirely on `HTTPSession.deinit` for proper cleanup, which ensures the correct sequence: (1) cancel all active requests, (2) invalidate session. This prevents crashes occurring inside APM libraries when they intercept network operations during session invalidation


parfeon:
LGTM!

PR #242, opened 2026-04-17
Fix message actions dropped when CryptoModule is enabled

jguz-pubnub:
fix: preserve message actions during history decryption

Fixed an issue where message actions were silently dropped from history responses when CryptoModule was enabled.


marcin-cebo (approved):

jguz-pubnub:
@pubnub-release-bot release as 10.1.4

PR #253, opened 2026-06-26
Harden Cryptor input validation, unify error surface, redact log secrets

jguz-pubnub:
fix: validate IV length and ciphertext block-alignment before decrypting

fix: return a uniform .decryptionFailure that never exposes the underlying CommonCrypto status

fix: guard LegacyCryptor random-IV path against payloads too short to contain an IV

fix: map encrypt-path errors to .encryptionFailure instead of .decryptionFailure 

fix: redact authKey and authToken in instance logs instead of emitting them in cleartext

fix: use .none as the default log level for the PubNubLogger public initializer to avoid an insecure default

docs: warn that verbose log levels may expose sensitive data and recommend long, high-entropy cipher keys

fix: make message action add/remove fire completion only once on error

refactor: make every PubNub instance manage its own presence-state container


Sources/PubNub/Logging/PubNubLogger.swift
```diff
@@ -86,10 +86,13 @@ public final class PubNubLogger {
...
   ///
+  /// - Warning: Enabling verbose levels such as `.debug`, `.trace`, or `.all` produces logs that may include  sensitive information,
+  /// such as API responses, user data, request URLs, authentication tokens, or internal system details. Avoid these levels in production builds.
+  ///
   /// - Parameters:
-  ///   - levels: The log levels to be included in the logger. Defaults to `.all`.
+  ///   - levels: The log levels to be included in the logger. Defaults to `.none`.
   ///   - writers: The writers to be used for logging. Defaults to the default log writers.
-  public init(levels: LogLevel = .all, writers: [LogWriter] = PubNubLogger.defaultLogWriters()) {
+  public init(levels: LogLevel = .none, writers: [LogWriter] = PubNubLogger.defaultLogWriters()) {
```
jguz-pubnub, line 95: The SDK always creates a PubNubLogger internally with the log level disabled. This change removes the insecure default from the public initializer.

marcin-cebo (approved):
LGTM!

jguz-pubnub:
@pubnub-release-bot release

PR #254, opened 2026-07-10
Add DataSync networking routers and endpoints

jguz-pubnub:
- Add DataSync routers for channels, users, entities, memberships, and relationships
- Extend `HTTPRouter` with the `Data Sync` service and DataSync-specific query keys
- Add error descriptions for empty DataSync ids, entity/relationship classes, and patch operations
- Only set the default `Content-Type` header when the request doesn't already provide one
- Add unit test coverage for the new DataSync routers

### DIFF
--- a/Sources/PubNub/Networking/Routers/DataSyncEntityRouter.swift
+++ b/Sources/PubNub/Networking/Routers/DataSyncEntityRouter.swift
@@ -179,8 +179,8 @@ struct DataSyncEntityRouter: HTTPRouter {
     switch endpoint {
     case let .all(entityClass, _, _, _, _, _, _):
       return isInvalidForReason((entityClass.isEmpty, ErrorDescription.emptyEntityClass))
-    case .create:
-      return nil
+    case let .create(body):
+      return isInvalidForReason((body.entityClass.isEmpty, ErrorDescription.emptyEntityClass))
     case let .fetch(id):
       return isInvalidForReason((id.isEmpty, ErrorDescription.emptyDataSyncId))
     case let .replace(id, _, _):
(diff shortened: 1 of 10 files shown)


Sources/PubNub/Networking/Routers/DataSyncEntityRouter.swift
```diff
@@ -0,0 +1,197 @@
...
+    return headers.compactMapValues { $0 }
+  }
+
+  // Validated
+  var validationErrorDetail: String? {
+    switch endpoint {
+    case let .all(entityClass, _, _, _, _, _, _):
+      return isInvalidForReason((entityClass.isEmpty, ErrorDescription.emptyEntityClass))
+    case .create:
+      return nil
```
marcin-cebo: Did you consider validation for empty class e.g.

```
 case let .create(body):
    return isInvalidForReason((body.entityClass.isEmpty, ErrorDescription.emptyEntityClass))
```
?

Sources/PubNub/Networking/Routers/DataSyncMembershipRouter.swift
```diff
@@ -0,0 +1,197 @@
...
+      break
+    }
+
+    return headers.compactMapValues { $0 }
+  }
+
+  var validationErrorDetail: String? {
+    switch endpoint {
+    case .all, .create:
+      return nil
```
marcin-cebo, line 183: Is this purposeful not to add validation for those operations?

marcin-cebo (approved):

Sources/PubNub/Networking/Routers/DataSyncEntityRouter.swift
jguz-pubnub: Makes sense! I'll implement it and add a test

Sources/PubNub/Networking/Routers/DataSyncMembershipRouter.swift
jguz-pubnub: This is justified here because all query parameters for Membership's get all endpoint are optional

PR #255, opened 2026-07-23
Add DataSync PAM token resources and create permission

jguz-pubnub:


Sources/PubNub/Helpers/PAMToken.swift
```diff
@@ -61,11 +61,28 @@ public struct PAMTokenResource: Codable, Equatable, Hashable {
   public let channels: [String: PAMPermission]
   public let groups: [String: PAMPermission]
```
marcin-cebo, line 62: Projections will be added in separate PR, right ?

marcin-cebo (approved):

Sources/PubNub/Helpers/PAMToken.swift
jguz-pubnub: Projections are part of the `meta` field which already exists [here](https://github.com/pubnub/swift/blob/10.1.8/Sources/PubNub/Helpers/PAMToken.swift#L22)

PR #256, opened 2026-07-29
Add DataSync subscribe event support

jguz-pubnub:
- Recognize the `e: 5` DataSync wire type, which previously fell back to  `.message` and silently surfaced DataSync changes as plain messages
- Add `PubNubDataSyncEvent` with entity/relationship created, updated and  deleted cases, delivered through the new `onDataSync` closure
- Add `PubNubDataSyncEntity` and `PubNubDataSyncRelationship` models, plus `PubNubDataSyncRemovedObject` for delete events that carry identity only
- Support both listener generations: `onDataSync` on `EventListenerInterface` and `didReceiveDataSyncEvent` on the legacy `SubscriptionListener`
- Parse the level-encoded `className` (`::name`, `:name:`, `name::`) into a name plus `PubNubDataSyncClassLevel`
- Decode date-only (`yyyy-MM-dd`) timestamps by adding `iso8601_dateOnly` as a further fallback after the full ISO 8601 and no-milliseconds formatters, so DataSync fields such as `expiresAt` parse correctly
- Fix `AnyJSON` dictionary equality, which compared subsets and so was asymmetric and inconsistent with `hash(into:)`
- Split `SubscribeRouterTests` into per-feature extensions to stay within the file-length limit
- Add the optional `entity_class` filter to the DataSync user and channel list routers, matching `GET /users` and `GET /channels` in `v4-data.yaml`


Documentation/PubNub_11_0_Migration_Guide.md
```diff
@@ -0,0 +1,73 @@
+# PubNub 11.0 Migration Guide
```
jguz-pubnub, line 1: I'm planning to release a major version for the following reasons:

1. Removing the `PubNubUser`, `PubNubSpace`, and `PubNubMembership` modules from SPM integration. They rely on AppContext under the hood and should never be advertised. I will remove them from the Package.swift file in another PR.
2. Removed local subscribe events to align with other SDKs. Subscribe events have been unified into the existing `PubNubEvent` type. Prior to this change, I would have had to copy-paste the same DataSync events to two nearly identical types: [PubNubEvent](https://github.com/pubnub/swift/blob/10.1.8/Sources/PubNub/Subscribe/API/PubNubEvent.swift) and [PubNubCoreEvent](https://github.com/pubnub/swift/blob/10.1.8/Sources/PubNub/Subscribe/Legacy/SubscriptionStream.swift#L79-L109). From now on, `PubNubEvent` is the single source of truth for all subscribe events.

Documentation/PubNub_11_0_Migration_Guide.md
```diff
@@ -0,0 +1,73 @@
...
+)
+```
+
+Both parameters have defaults (`.main` and `true`), so `SubscriptionListener()` and `SubscriptionListener(queue:)` continue to compile unchanged — only post-construction assignment is affected. If you subclass `BaseSubscriptionListener` or `PubNubEntityListener`, forward these values through `super.init(queue:supressCancellationErrors:)` rather than setting the properties in your own initializer.
+
+## Deprecations (Non-Breaking Changes)
+
+### Catch-all Event Callbacks
+
+The callbacks that deliver the whole `PubNubEvent` stream are deprecated in favor of the granular, per-type callbacks. They still work in 11.0 but will be removed in a future release:
```
marcin-cebo, line 68: What is a motivation for deprecating delivery of the whole PubNubEvent stream?

Sources/PubNub/Models/PubNubDataSync.swift
```diff
@@ -0,0 +1,133 @@
...
+//
+
+import Foundation
+
+/// The level a DataSync class is registered at
+public enum PubNubDataSyncClassLevel: Hashable {
+  /// A class shared across all accounts
+  case global
+  /// A class scoped to the account
+  case account
```
marcin-cebo, line 18: Does server provide account level?

Sources/PubNub/Models/PubNubDataSync.swift
```diff
@@ -0,0 +1,133 @@
...
+  /// The date the entity expires, derived from the time-to-live of its class
+  public let expiresAt: Date
+  /// The entity status
+  public let status: String?
+  /// The entity fields
+  public var payload: JSONCodable? { concretePayload }
+
+  let concretePayload: AnyJSON?
+
+  init(
```
marcin-cebo, line 50: This is internal and PubNubDataSyncRemovedObject has public init. Is this intentional?

marcin-cebo (approved):

Documentation/PubNub_11_0_Migration_Guide.md
jguz-pubnub: This is to stay aligned with other SDKs that don't have this concept. Also, this kind of closure causes extremely long switch cases (here's a good [example](https://www.pubnub.com/docs/sdks/swift/api-reference/publish-and-subscribe#sample-code-13)) that most static code analysis tools, like SwiftLint, report as violations due to high cyclomatic complexity.

Sources/PubNub/Models/PubNubDataSync.swift
jguz-pubnub: Good catch, `PubNubDataSyncRemovedObject` should have an internal initializer.

PR #257, opened 2026-08-05
Add classLevel to DataSync subscribe event metadata

jguz-pubnub:


Sources/PubNub/Networking/Routers/Subscribe Payloads/SubscribeDataSyncPayload.swift
```diff
@@ -64,9 +65,9 @@ extension SubscribeDataSyncPayload: Decodable {
 
     let action = try metadata.decode(Action.self, forKey: .event)
     let type = try metadata.decode(ObjectType.self, forKey: .type)
-    let rawClassName = try metadata.decode(String.self, forKey: .className)
+    let className = try metadata.decode(String.self, forKey: .className)
+    let classLevel = PubNubDataSyncClassLevel(stringValue: try metadata.decode(String.self, forKey: .classLevel))
```
marcin-cebo, line 69: I seems that when classLevel is absent this code will throw. So we assume classLevel is always there?

Sources/PubNub/Models/PubNubDataSync.swift
```diff
@@ -18,6 +18,35 @@ public enum PubNubDataSyncClassLevel: Hashable {
...
+      return "Account"
+    case .subKey:
+      return "SubKey"
+    case let .unknown(rawValue):
+      return rawValue
+    }
+  }
+
+  init(stringValue: String) {
+    switch stringValue {
```
marcin-cebo, line 39: Is seems that we are sure that classLevel is exactly of shape  "Global", "Account", "SubKey" not "GLOBAL" or "global" etc.

Sources/PubNub/Models/PubNubDataSync.swift
```diff
@@ -18,6 +18,35 @@ public enum PubNubDataSyncClassLevel: Hashable {
...
+  /// A level this version of the SDK doesn't recognize, along with the value received from the server
+  case unknown(String)
+
+  /// The value used to represent the level on the wire
+  public var stringValue: String {
+    switch self {
+    case .global:
+      return "Global"
+    case .account:
+      return "Account"
```
marcin-cebo, line 30: I am wodering what would happen when you create Class on Account level and on SubKey level then you want to create instance. Current API createEntity doesn't allow to specify entityLevel.

marcin-cebo (approved):

Sources/PubNub/Networking/Routers/Subscribe Payloads/SubscribeDataSyncPayload.swift
jguz-pubnub: Yes, this is the contract we rely on.

Having a dedicated `classLevel` field is much better than parsing a `className` with `:` or `::` as indicators of class level.

Sources/PubNub/Models/PubNubDataSync.swift
jguz-pubnub: Good catch. From what I see, creating an account-level entity raises an error and is not supported. This would explain the lack of `entityLevel` for the createEntity endpoint.

Sources/PubNub/Models/PubNubDataSync.swift
jguz-pubnub: Yes, here's what I found in their code:

```
enum class ClassLevel {
   Global, Account, SubKey
}
```

Sources/PubNub/Models/PubNubDataSync.swift
```diff
@@ -18,6 +18,35 @@ public enum PubNubDataSyncClassLevel: Hashable {
...
+    case .subKey:
+      return "SubKey"
+    case let .unknown(rawValue):
+      return rawValue
+    }
+  }
+
+  init(stringValue: String) {
```
marcin-cebo, line 38: Is this ok that init is not public?

marcin-cebo (approved):

Sources/PubNub/Models/PubNubDataSync.swift
jguz-pubnub: Yes, we don't want callers to instantiate this type themselves.

PR #262, opened 2026-08-28
Add APNs push type to PubNubPushConfig

jguz-pubnub:
feat: add APNs push type to PubNubPushConfig

Add an optional `pushType` property that maps to the `apns-push-type` header required for non-alert notifications such as VoIP. When omitted, the payload excludes the key, and APNs defaults to alert.


Sources/PubNub/Models/PubNubAPNSPayload.swift
```diff
@@ -26,23 +26,32 @@ public struct PubNubPushConfig: Codable, Hashable {
...
+  /// The type of the remote notification
+  ///
+  /// This will populate the APN's `apns-push-type` header.
+  ///
+  /// When `nil` the key is omitted from the payload, and PubNub defaults the header to ``PubNub/PushType/alert``.
+  public let pushType: PubNub.PushType?
 
   enum CodingKeys: String, CodingKey {
     case authMethod = "auth_method"
     case collapseID = "collapse_id"
```
marcin-cebo, line 38: Is seems that `collapseID` is encoded as `collapse_id` but in doc:
https://www.pubnub.com/docs/general/push/ios
there is: `pn_collapse_id`

Sources/PubNub/Models/PubNubAPNSPayload.swift
```diff
@@ -26,23 +26,32 @@ public struct PubNubPushConfig: Codable, Hashable {
   ///
   /// This will populate the APN's `apns-expiration` header
   public let expiration: Date?
+  /// The type of the remote notification
+  ///
+  /// This will populate the APN's `apns-push-type` header.
+  ///
+  /// When `nil` the key is omitted from the payload, and PubNub defaults the header to ``PubNub/PushType/alert``.
+  public let pushType: PubNub.PushType?
```
marcin-cebo, line 34: Doesn't new pushType clash with pushType from listPushChannelRegistrations.

Maybe APNSPushType could solve the clash ?

marcin-cebo (approved):

Sources/PubNub/Models/PubNubAPNSPayload.swift
jguz-pubnub: There's no clash. I preserved the existing `PubNub.PushService` type that describes notification providers like APNS or FCM, and changed the inner argument name to `pushService`.

Sources/PubNub/Models/PubNubAPNSPayload.swift
marcin-cebo: In JS: https://www.pubnub.com/docs/sdks/javascript/api-reference/mobile-push
there is:
`"collapse_id": "invitations",`
so maybe there is an issue in Swift doc and `collapse_id` should be used in code and correction should cover only doc? I am not sure.

jguz-pubnub: I checked it using an example app - `collapse_id` is correct

jguz-pubnub:
@pubnub-release-bot release as 10.2.0

PR #263, opened 2026-09-01
Rename DataSync GET filtering parameters and require expiresAt

jguz-pubnub:
- `filter` → `filter_fast` (`filterFast:`), 
-  `filter_advanced` → `filter` (`filter:`)
-  `expiresAt` becomes non-optional
- enable DataSync integration tests
- add scripts to show how DataSync classes that integration tests rely on were created


Scripts/DataSync/create-classes.sh
jguz-pubnub, line 1: This file is not part of any target and will not be included in the resulting binary. It's kept in the repository to show how DataSync classes that integration tests rely on were created.

Sources/PubNub/APIs/DataSync+PubNub.swift
```diff
@@ -133,8 +133,8 @@ public extension PubNub.DataSyncAPI {
   ///   - classLevel: The level to resolve the class name at. If omitted, the first level declaring it wins, searching `subKey`, `account`, then `global`
   ///   - cursor: The ``PubNubDataSyncPage/cursor`` of a previous page, or `nil` for the first page
   ///   - limit: The number of entities to retrieve, between 1 and 100. Defaults to 20
-  ///   - filter: Expression used to filter the results. Mutually exclusive with `filterAdvanced`
-  ///   - filterAdvanced: Advanced expression used to filter the results. Mutually exclusive with `filter`
+  ///   - filterFast: Filters the latest data, including changes made immediately before the request. Supports a limited number of filter conditions. Mutually exclusive with `filter`
+  ///   - filter: Supports more flexible filtering, but recent changes may take a short time to appear in the results. Mutually exclusive with `filterFast`
```
marcin-cebo, line 136: Here is description of filterFast, filter, sort. What do you think about adding info about what fields can be used for filter/search ?

     * @param filterFast Optional filter expression. Filtering is only allowed on the entity class's properties
     *   whose filtering mode is not disabled (i.e. those the class marks as filterable); filtering on any
     *   other property returns a server error. The filterable/sortable set has no fixed default — it is
     *   whatever the (required) [className] declares. In addition, the built-in fields `id`, `createdAt`,
     *   `updatedAt`, and `status` (case-sensitive, exactly as spelled) are always filterable and sortable on
     *   any class, regardless of its declared properties. `filterFast` is
     *   strongly consistent — it always reflects the latest writes — but accepts fewer conditions than
     *   [filter]; a limit on the number of conditions applies and can be adjusted by PubNub support
     *   (see the PubNub DataSync documentation for the current limit). For larger or more complex queries use
     *   [filter]. At most one of [filterFast] and [filter] may be supplied; sending both is
     *   rejected with an error.
     * @param filter Optional advanced filter expression. Uses the same syntax and filterable-property
     *   rules as [filterFast] and is not subject to the same condition limit, so use it for larger or more complex
     *   queries. The trade-off is consistency: `filter` is eventually consistent (recent writes may not
     *   yet be reflected), whereas [filterFast] is strongly consistent. At most one of [filterFast] and [filter]
     *   may be supplied; sending both is rejected with an error.
     * @param sort Optional sort criteria applied in order; each [PNDataSyncSortField] sorts on a payload
     *   property either ascending (default) or descending. Sorting is governed by the same filterable-property
     *   rule as [filterFast] — the properties the (required) [className] marks as filterable, plus the built-in
     *   fields `id`, `createdAt`, `updatedAt`, and `status`.


For getChannels it looks like this:

     * @param filterFast Optional filter expression. Filtering is only allowed on the entity class's properties
     *   whose filtering mode is not disabled (i.e. those the class marks as filterable); filtering on any
     *   other property returns a server error. For the default `Channel` class this set is `name` and `type`,
     *   but a custom class or subclass may declare additional filterable properties. These are filterable
     *   indexes over `/payload/name` and `/payload/type` — both nullable, not a required or exclusive payload
     *   schema; the `payload` stays arbitrary JSON. In addition, the built-in fields `id`, `createdAt`,
     *   `updatedAt`, and `status` (case-sensitive, exactly as spelled) are always filterable and sortable on
     *   any class, regardless of its declared properties. `filterFast` is strongly consistent — it always reflects the
     *   latest writes — but accepts fewer conditions than [filter]; a limit on the number of conditions
     *   applies and can be adjusted by PubNub support (see the PubNub DataSync documentation for the current
     *   limit). For larger or more complex queries use [filter]. At most one of [filterFast] and
     *   [filter] may be supplied; sending both is rejected with an error.
     * @param filter Optional advanced filter expression. Uses the same syntax and filterable-property
     *   rules as [filterFast] and is not subject to the same condition limit, so use it for larger or more complex
     *   queries. The trade-off is consistency: `filter` is eventually consistent (recent writes may not
     *   yet be reflected), whereas [filterFast] is strongly consistent. At most one of [filterFast] and [filter]
     *   may be supplied; sending both is rejected with an error.
     * @param sort Optional sort criteria applied in order; each [PNDataSyncSortField] sorts on a payload
     *   property either ascending (default) or descending. Sorting is governed by the same filterable-property
     *   rule as [filterFast] (default `Channel` class: `name` and `type`, plus the built-in fields `id`,
     *   `createdAt`, `updatedAt`, and `status`).

fastlane/Fastfile
```diff
@@ -51,10 +51,9 @@ lane :integration_test do
   secret_key = ENV['SDK_SECRET_KEY']
```
marcin-cebo: FYI in DataSyncEntityEndpointIntegrationTests tests:
testGetEntitiesPinnedToAClassVersionReturnsOnlyThatVersion
testGetEntitiesWithoutAClassVersionReturnsEveryVersion
testGetEntitiesFollowsCursorToSecondPage 

still filter is passed and now it has different meaning because after rename it points to what used to by name filter_advance 

The thing is that once the server change is implemented this filter would hit OpenSearch that has letancy between write and read and calling listEntities could not return all results

Sources/PubNub/APIs/DataSync+PubNub.swift
```diff
@@ -133,8 +133,8 @@ public extension PubNub.DataSyncAPI {
   ///   - classLevel: The level to resolve the class name at. If omitted, the first level declaring it wins, searching `subKey`, `account`, then `global`
   ///   - cursor: The ``PubNubDataSyncPage/cursor`` of a previous page, or `nil` for the first page
   ///   - limit: The number of entities to retrieve, between 1 and 100. Defaults to 20
-  ///   - filter: Expression used to filter the results. Mutually exclusive with `filterAdvanced`
-  ///   - filterAdvanced: Advanced expression used to filter the results. Mutually exclusive with `filter`
+  ///   - filterFast: Filters the latest data, including changes made immediately before the request. Supports a limited number of filter conditions. Mutually exclusive with `filter`
```
marcin-cebo, line 135: Is this possible to use filterFast on properties that are defined as 
 "filtering": "simple"    ?

I am wondering if this is something that should be added to inline doc?

marcin-cebo (approved):

Sources/PubNub/APIs/DataSync+PubNub.swift
jguz-pubnub: It should be possible. On second though, the whole filter feature looks quite complex. I don't think we're able to explain its every aspect in inline docs. It has to be balanced

Sources/PubNub/APIs/DataSync+PubNub.swift
jguz-pubnub: I expect inline comments to stay short and concise, and I don't think we can explain every aspect of the filtering feature. This should be general documentation covering all SDKs, or a per-SDK subpage explaining how to work with this feature for a certain language. 

This isn't a new approach. For example, the existing `allUserMetadata` method briefly describes its filter parameter and links to external documentation that covers the functionality in detail. You can find it below:

https://github.com/pubnub/swift/blob/10.1.8/Sources/PubNub/APIs/Objects%2BPubNub.swift#L407

